PR #2, opened 2022-07-21
Create bind_tables and add test suite

rsh52:
# Description
This PR introduces the `bind_tables()` function which takes a `read_redcap_tidy()` output and assigns the data contents (via `$redcap_data`) to a specified environment output. By default, all data tables are assigned to their `$redcap_form_name`s and loaded into the `global_env()`. Users can optionally specify a different environment using `rlang::new_environment()`, or specify specific forms or structures to load in the `redcap_form_name` and `structure` arguments. 

**Note:** I am unsure how to test this in a `shiny` reactive context, let me know if you have any thoughts on this!

# Proposed Changes 

- Create `bind_tables()`
- Create associated test suite for different argument combinations
- Update `pkgdown` documentation

### Issue Addressed
NA

### PR Checklist

- [NA] Issues linked to the PR using [GitHub's list of keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

# Code Review

### Code Review Checklist


tests/testthat/test-bind_tables.R
```diff
@@ -0,0 +1,49 @@
+# Load initial variables
```
skadauke, line 1: This test suite is great, but instead of relying on `read_redcap_tidy()` to create the data set to test, it would be better to load a saved .rds file that contains the output of `read_redcap_tidy()`.

tests/testthat/test-bind_tables.R
```diff
@@ -0,0 +1,49 @@
...
+
+  read_redcap_tidy(redcap_uri, token) %>%
+    bind_tables(environment = sample_env)
+
+  expect_true(exists("nonrepeated", envir = sample_env))
+  expect_true(exists("nonrepeated2", envir = sample_env))
+  expect_true(exists("repeated", envir = sample_env))
+})
+
+test_that("bind_tables works with forms specification", {
```
skadauke, line 27: The way this test is written, it tests for both the `environment` and `redcap_form_name` specification. Unit tests should really only test one behavior in isolation so I would use `global_env` not `sample_env` for this test.

Also, I would actually test whether you can pass `c("repeated", "nonrepeated")` to `redcap_form_name` to make sure passing multiple form names works as expected.

tests/testthat/test-bind_tables.R
```diff
@@ -0,0 +1,49 @@
...
+  read_redcap_tidy(redcap_uri, token) %>%
+    bind_tables(environment = sample_env, redcap_form_name = "nonrepeated")
+
+  expect_true(exists("nonrepeated", envir = sample_env))
+  expect_false(exists("nonrepeated2", envir = sample_env))
+  expect_false(exists("repeated", envir = sample_env))
+})
+
+test_that("bind_tables works with structure specification", {
+
```
skadauke, line 40: As above, would use `global_env`

tests/testthat/test-bind_tables.R
```diff
@@ -0,0 +1,49 @@
+# Load initial variables
+token <- Sys.getenv("REDCAPTIDIER_LONGITUDINAL_API")
+redcap_uri <- Sys.getenv("REDCAP_URI")
+
+test_that("bind_tables works with no specifications", {
+
+  read_redcap_tidy(redcap_uri, token) %>%
+    bind_tables()
+
+  expect_true(exists("nonrepeated", envir = global_env()))
```
skadauke, line 10: Is `envir = global_env()` necessary here?

skadauke (commented):
This looks great! Some suggestions for tightening the test suite.

tests/testthat/test-bind_tables.R
skadauke, line 10: Since the spec is that `bind_tables()` is that it `plop`s data frames into an environment, I think at least for the case in which the environment is the global environment, you could use `expect_s3_class(x, "data.frame")` here.

tests/testthat/test-bind_tables.R
rsh52: Done 👍

tests/testthat/test-bind_tables.R
rsh52: You're right, default behavior is to check global anyway. Removed 👍

rsh52: Added 3 tests for the 3 dataframes expected 👍

tests/testthat/test-bind_tables.R
rsh52: I wrote the test to that it checks `c()` and it does still work 👍 

I'm having an issue with using `global_env()` instead of `sample_env()`, the tests wind up keeping the data objects between tests which results in failures. I looked into `withr::with_environment` but couldn't get it working, for now I added in a `rm()` line at the end of the tests to remove the expected objects but realize this is probably not good practice.

skadauke: I think your solution is fine... nit: would add a comment before the `rm()` line, like so:
```
# Clean up
rm(...)
```
Also, I don't think you need to specify `global_env()`, do you?

rsh52: Just tried it on my home Windows machine, and had this happen on the Mac as well. Without specifying the `global_env()`, objects still get loaded into it and the tests wind up failing. I can get around this by adding `envir = global_env()` to the `rm()` statements but then it's just a preference of where you want it specified. Unsure why this happens, since the default is to use the "active" directory and `identical(global_env(), current_env())` is `TRUE`.

Maybe it has to do with the `testthat` environment, but I don't really know.

skadauke (approved):
Looks good to me!

skadauke (approved):

PR #21, opened 2022-08-17
Label Support for Users

### LINKED ISSUE #8: read_redcap_tidy should support using labels instead of raw values
# Feature Request Description

Currently, the tibbles that `read_redcap_tidy` generates contain the raw values (usually but not always serial integer numbers) for  the following field types:

- dropdown-single
- radio-single
- checkbox-multiple
- yesno
- truefalse

There should be a switch to `read_redcap_tidy` (set to `TRUE` by default) to make the values contain the labels instead of raw values. 

# Proposed Solution

For the following field types, the labels are trivial and can be easily implemented programmatically:

- yesno: 0 -> `no` / 1 -> `yes` (factor)
- truefalse: 0 -> `FALSE` / 1 -> `TRUE` (logical)
- checkbox-multiple: 0 -> `FALSE` / 1 -> `TRUE` (logical)

For the following types, the mapping of raw to label can be determined using the `parse_labels` function.  

- dropdown-single
- radio-single

An important implementation detail here is that the order of the choices may have meaning. For this reason, these categorial field types should be coded as `factor` in the resulting tibble, with the levels corresponding to the order of choices.

### Checklist

- [X] The issue is atomic
- [X] The issue description is documented
- [X] The issue title describes the problem succinctly
- [X] Developers are assigned to the issue
- [X] Labels are assigned to the issue

rsh52:
# Description
This PR introduces the option for users to select between `raw` or `label` outputs from `read_redcap_tidy` which impacts multichoice field types (truefalse, yesno, checkboxes, radio buttons, and dropdown lists). The current default selection is `label`s.

To address this, I made a new function, `multi_choice_to_labels` (name can be changed), which is used in the overarching `read_redcap_tidy` function if `raw_or_label = "label"`. From there, the function `mutate`s yesno/truefalse/checkbox outputs to those specified in #8 using a `for` loop with a series of conditionals checked against the metadata `field_type`. 

Later in the loop, `parse_labels` is used on `radio` and `dropdown` `field_type`s to return label data. `factor`s are enforced and orders are specified in the order they are output by `parse_labels`. 

> **Note:** In the latter portion, I made use of `rlang` injections which I believe could be a little more slick. I could not get my `left_join(by = '')` to work well with this and made a comment about it.

# Proposed Changes 

- Create function `multi_choice_to_labels`
- Create preliminary test suite to ensure `multi_choice_to_labels` returns expected data types
- Remove `raw_or_label` calls from functions where no longer necessary

### Issue Addressed
**Closes #8**

### PR Checklist

# Code Review

### Code Review Checklist


tests/testthat/test-clean_redcap.R
```diff
@@ -45,8 +45,8 @@ test_that("extract_nonrepeat_table tibble contains expected columns and data typ
   expect_character(out$text)
```
skadauke, line 45: I think you have to repeat this section on a data set in which `raw_or_label == label`.

tests/testthat/test-clean_redcap.R
```diff
@@ -45,8 +45,8 @@ test_that("extract_nonrepeat_table tibble contains expected columns and data typ
...
   expect_character(out$note)
   expect_double(out$calculated)
-  expect_double(out$dropdown_single)
-  expect_double(out$radio_single)
+  expect_character(out$dropdown_single)
```
skadauke, line 48: Assuming this is for `label` - shouldn't `dropdown_single` and `radio_single` be a `factor` rather than `character`?

tests/testthat/test-clean_redcap.R
```diff
@@ -45,8 +45,8 @@ test_that("extract_nonrepeat_table tibble contains expected columns and data typ
...
+  expect_character(out$radio_single)
   expect_double(out$checkbox_multiple___1)
```
skadauke, line 50: Assuming this is for `label` - shouldn't `checkbox` and `truefalse` be `logical`? And `yesno` should be `factor`?

skadauke (commented):
Hey - I'm not sure this is right yet, form looking at the test suite. Did you look at the output of running the classic test DB with `raw_or_label = label`? We can look at this tomorrow.

tests/testthat/test-clean_redcap.R
rsh52: Yes, I misread the issue and thought factor was only for the latter two (dropdown/radio). I can add an a quick coercion to the function that coerces to `factor`, I actually had it in an earlier commit and removed it.

tests/testthat/test-clean_redcap.R
rsh52: Yes, I'll look into this tomorrow.

rsh52:
> Hey - I'm not sure this is right yet, form looking at the test suite. Did you look at the output of running the classic test DB with `raw_or_label = label`? We can look at this tomorrow.

Yes, I have to re-address these tests. The behavior is working as expected though in the call below, but `yesno` needs to be fixed to be `logical` instead of `character`:

```r
read_redcap_tidy(redcap_uri = Sys.getenv("REDCAP_URI"),
                 token = Sys.getenv("REDCAPTIDIER_CLASSIC_API"),
                 raw_or_label = "label") %>% 
  bind_tables()
```

Where:

```r
> data_field_types$radio_single
[1] B    C    <NA>
Levels: A < B < C
```

```r
> data_field_types$dropdown_single
[1] one   three <NA> 
Levels: one < two < three
```

```r
> class(data_field_types$checkbox_multiple_2___3dddd)
[1] "logical"
```

```r
> class(data_field_types$truefalse)
[1] "logical"
```

tests/testthat/test-clean_redcap.R
rsh52: Actually I think this is correct here, I changed the content of the questions in the sample REDCap to have non-numeric labels and raw data (both sides of the comma), and `clean_redcap` does not enforce any label manipulation or cross-referencing with the metadata. That's currently done at the top level `read_redcap_tidy` separate from `clean_redcap`. So here I just needed to update what we expect to come out of REDCap itself, but reordering of things might be necessary.

tests/testthat/test-clean_redcap.R
rsh52: See above, these tests are based on `clean_redcap` and not the output of the new function. However, I did update the output of `yesno` in the newest commit.

rsh52:
Hm, seems the R-CMD-Check is stuck and it won't let me cancel the workflow... Will check later in the day but it does work locally.

R/utils.R
```diff
@@ -150,3 +144,64 @@ update_field_names <- function(db_metadata, raw_or_label = 'raw'){
...
+#'
+#' @param db_data A REDCap database object
+#' @param db_metadata A REDCap metadata object
+#'
+#' @import dplyr
+#' @keywords internal
+
+multi_choice_to_labels <- function(db_data, db_metadata){
+
+  for (i in 1:nrow(db_metadata)) {
```
skadauke, line 157: I would put the following ahead of the `for` loop and then delete lines 162-184.

```r
# Handle columns where we change 0/1 to FALSE/TRUE (logical)
logical_cols <- db_metadata %>%
  filter(field_type %in% c("yesno", "truefalse", "checkbox")) %>%
  pull(field_name_updated)
 
db_data2 <- db_data %>%
  mutate(across(.cols = logical_cols, as.logical))
```

R/utils.R
```diff
@@ -150,3 +144,64 @@ update_field_names <- function(db_metadata, raw_or_label = 'raw'){
...
+    if (db_metadata$field_type[i] == "checkbox") {
+      db_data <- db_data %>%
+        mutate(
+          across(.cols = field_name, .fns = ~case_when(. == 1 ~ TRUE, . == 0 ~ FALSE, TRUE ~ NA))
+        )
+    }
+
+    # dropdown and radio datatype handling ----
+    if (db_metadata$field_type[i] == "dropdown" | db_metadata$field_type[i] == "radio") {
+
```
skadauke, line 188: better: `db_metadata$field_type[i] %in% c("dropdown", "radio")`

R/utils.R
```diff
@@ -150,3 +144,64 @@ update_field_names <- function(db_metadata, raw_or_label = 'raw'){
...
+      # Retrieve parse_labels key for given field_name
+      parse_labels_output <- parse_labels(db_metadata$select_choices_or_calculations[i])
+
+      # Replace values from db_data$(field_name) with label values from  parse_labels key
+      db_data <- db_data %>%
+        left_join(parse_labels_output %>% rename(!!field_name := raw), # Could not get working in by argument, instead inject field name for rename in parse_labels_output
```
skadauke, line 194: I don't think the `rename` is necessary if you use a named `by` -- i.e. `by = c("raw" = field_name)` or similar

R/utils.R
```diff
@@ -150,3 +144,64 @@ update_field_names <- function(db_metadata, raw_or_label = 'raw'){
...
+                  by = field_name) %>%
+        mutate(!!field_name := label) %>% # Again, use rlang var injection
+        select(-label)
+
+      db_data <- db_data %>%
+        mutate(
+          across(.cols = field_name, .fns = ~factor(., ordered = T, levels = parse_labels_output$label))
```
skadauke, line 201: I don't think we want to make `ordered = TRUE` the default.

R/utils.R
rsh52: Just want to make sure, you mean `db-data <- db_data` not `db_data2 <- db_data` right? Accidentally blindly copy/pasted and had test fails because we aren't using `db_data2` elsewhere haha

rsh52:
Changes implemented, checks passing (hooray!). Just need to resolve the new unrelated bug 😅

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,28 @@
...
+
+  # Test general structure
+  expect_true(is.data.frame(out))
+  expect_true(nrow(out) > 0)
+
+  # Test multichoice options return expected values and datatypes
+  expect_logical(out$yesno)
+  expect_logical(out$truefalse)
+
+  checkbox_col_types <- map(out %>% select(contains("checkbox")), class)
```
skadauke, line 17: Ugghhhh... can you make this less complicated? I really don't know what's going on here. There should really be no `for` loops or `map` functions in tests.

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,28 @@
...
+  checkbox_col_types <- data.frame(
+    matrix(unlist(checkbox_col_types),
```
skadauke, line 19: ???

skadauke (commented):
See comments about `test-utils.R`. Also, I think it would be good to address the bug(s) we discovered today inside of this PR.

tests/testthat/test-utils.R
rsh52: Haha I knew you'd call this out. I was trying to make an elegant way of capturing all `checkbox_*` fields to test, but if you prefer we can either test one field or do many single tests. I think in the `clean_*` tests we just tested a single `checkbox_*` field.

tests/testthat/test-utils.R
rsh52: See above haha

rsh52:
Updated `parse_labels` with the comments we made today over email and added a few tests for string expectations. The function will now look for the first comma in splits on `|`s, and will fail if:

- supplied matrix dimensions are not even
- split objects contain no commas

It will still fail silently though for the example you wrote in the red/green email:
`raw, label | continued label, with commas, and pipe | another, one ` 

Where it is both even and contains commas. Unsure how to tackle this one.

Also sadly calling Prodigy and CART still fails.. 😞  All of our test REDCap's pass though. Will have to look further into this on Monday.

rsh52:
Brief update: The error related to CART/Prodigy turned into 4 errors, but I figured out and resolved 3 of them! Unfortunately I need to take the dog out and leave soon but I have all of the resolutions documented to propose when I get the last bug squashed. Since I'm running out the door, here are my jumbled notes:

> There is still one error related to Prodigy failing on raw/label matrix structure when evaluating labels.
 
## **Prodigy/CAR-T Error Resolution (label):**

When radio/dropdown values are assigned raw values of integers, REDCap outputs them as integers.
 
image

Solution is to coerce these values to character inside multi_choice_to_labels so left_join doesn’t fail.

## **Prodigy Error Resolution (raw):**
image

Metadata exports were still including “descriptive” field_types, which don’t come out in the database. So email variables were failing because there was nothing to select on.
 
Quick fix is to just filter out all “descriptive” field types as soon as metadata is called.
 
## **CAR-T Export Error Resolution (raw):**
image

This one is pretty funny, I was baffled why these weren’t coming out and they must be new additions to the database. Checked my user rights and saw that I did not have full dataset access to these instruments specifically:
image
image

rsh52:
Alright, found the last error was related to a descriptive text field in Prodigy with field embedding that was accidentally made as a radio field. The `parse_labels` failed because the `select_choices_*` variable was empty where a string was expected. I think we're good now! 🤞

R/utils.R
```diff
@@ -69,15 +69,31 @@ link_arms <- function(
...
 
-parse_labels <- function(string, raw_or_label){
+parse_labels <- function(string){
   out <- string %>%
-    strsplit(" \\| |, ") %>% # split either by ' | ' or ', '
+    strsplit(" \\| ") # Split by "|"
+
+  # Check there is a comma in all | delimited strsplit elements
+  if (!all(grepl(",", out[[1]]))) {
+    stop(paste0("The raw/label structure of the string is not analyzable: ", string))
```
skadauke, line 82: Do you want to use `checkmate::assert` here? 

Also, I would change the message to: "Cannot parse the select_choices_or_calculations field from REDCap metadata. This may happen if there is a pipe character \`|\` inside the label"

R/utils.R
```diff
@@ -69,15 +69,31 @@ link_arms <- function(
...
+  }
+
+  # split on the _first_ comma in each element
+  out <- out %>%
     unlist() %>%
+    stri_split_fixed(pattern = ", ", n = 2) %>% # Split by first ","
+    unlist()
+
+  # Check if vector is even for matrix creation. If not, then fail.
+  if (length(out) %% 2 != 0) {
```
skadauke, line 92: Same comment as above, would use the same error message

tests/testthat/test-utils.R
```diff
@@ -0,0 +1,42 @@
...
+  expect_logical(out$checkbox_multiple___1)
+  expect_logical(out$checkbox_multiple_2___4eeee5)
+
+  expect_factor(out$dropdown_single)
+  expect_equal(levels(out$dropdown_single), c("one", "two", "three"))
+  expect_factor(out$radio_single)
+  expect_equal(levels(out$radio_single), c("A", "B", "C"))
+})
+
+test_that("parse_labels works", {
```
skadauke, line 25: This is my favorite test of all tests you've written! 🥇

skadauke (commented):
This looks great! 

I think we should gracefully deal with the situation when you have a parse_labels receives an empty string. It's not technically illegal in REDCap for a radio button or similar field to have no categories so I wouldn't cause failure here. But it's probably an error in the database so the sensible thing would be to throw a warning "Warning: The field {field_name} in {instrument_name} is a {field_type}, however it does not have any categories." I would add this into the current PR.

For the issue you encountered in the CART DB, I think we should make a separate Github issue. I think we should detect when not all of the tables or fields that we expect from the metadata are present. I think this should not fail, but there should be a warning that informs the analyst of which instruments are missing, and why that might be the case. It should be possible to suppress this warning as it may be intended if a database is set up to export only specific instruments and/or identifier fields are removed (what is the difference between "De-identified* and "Remove All Identifier Fields"??).

R/utils.R
rsh52: I tried to initially but couldn't figure out how to get a custom error message out of `assert`/`check`. The default when I used it wasn't helpful for the user., i.e. `Error: Assertion on 'all(grepl(",", out[[1]]))' failed: Must be TRUE.`

I would still like to keep the `string` output because I think it can be helpful to the user to locate what field is causing the issue. Otherwise agree, implementing now.

tests/testthat/test-utils.R
rsh52: What high praise! Great way to start the weekend 🎉

rsh52:
> This looks great!
> 
> I think we should gracefully deal with the situation when you have a parse_labels receives an empty string. It's not technically illegal in REDCap for a radio button or similar field to have no categories so I wouldn't cause failure here. But it's probably an error in the database so the sensible thing would be to throw a warning "Warning: The field {field_name} in {instrument_name} is a {field_type}, however it does not have any categories." I would add this into the current PR.

🎉 🎉 🎉 

I agree, I can add this. In a way it's almost good I made some mistakes in REDCap 😆 

> For the issue you encountered in the CART DB, I think we should make a separate Github issue. I think we should detect when not all of the tables or fields that we expect from the metadata are present. I think this should not fail, but there should be a warning that informs the analyst of which instruments are missing, and why that might be the case. It should be possible to suppress this warning as it may be intended if a database is set up to export only specific instruments and/or identifier fields are removed (what is the difference between "De-identified* and "Remove All Identifier Fields"??).

Will get on writing this up, not sure how to make it not fail with the current setup. Unless we implement behavior that also removes this missing data from the metadata so there's no crash by the time `left_join` comes around? In that case I think a warning would be good.

IIRC I think de-identified/remove identifiers is a setting in REDCap that labels a variable as an identifier so would block it from export. It may also date-shift dates but I'm not sure, would have to test.

rsh52:
@skadauke I created two more issues for `form_status_complete` and to address missing instrument handling (such as what I experienced with the API).

In the most recent commit here I added a `warning` check to the `multi_choice_to_labels` to check for empty `select_choices_or_calculations` strings using the message you provided. I also wrapped the existing error checks to pass if this is met (I made it a nested loop for easier reading). I also made two more tests to catch the behavior.

Let me know what you think and if we can go ahead and merge this.

skadauke (commented):
Looks great to me!

skadauke (approved):

PR #32, opened 2022-09-02
checkbox_appender Correctly Handles "-"s

### LINKED ISSUE #31: [BUG] redcap_ready_tidy fails on checkbox field with negative value
# Expected Behavior
`redcap_read_tidy` should be able to handle all types of fields

# Current Behavior
I have several checkbox fields in my project where the value `'Unknown'` is coded as `-99`. In the data export REDCap converts the minus sign `-` into an underscore `_`. For example, if the patients identifies as white then `race___4` will be `1` but if they refuse to supply their race then `race____99` will be `1` -- notice the four underscores instead of the 3 underscores for the `'Unknown'` option

It seems like this break assumptions that `redcap_ready_tidy` makes, since I get this error message:

```r
> read_redcap_tidy(uri, token)
```
```
Error in `mutate()`:
! Problem while computing `..1 = across(.cols =
  all_of(logical_cols), as.logical)`.
Caused by error in `across()`:
! Can't subset columns that don't exist.
x Column `race___-99` doesn't exist.
```

After searching through the source code I suspect the underlying issue is here:

https://github.com/CHOP-CGTDataOps/REDCapTidieR/blob/1fd9b7763fbd85e59be7a0a5659092c8ab1d7dfb/R/utils.R#L133-L142

### Checklist

- [ ] The issue is atomic
- [X] The issue description is documented
- [X] The issue title describes the problem succinctly
- [ ] Developers are assigned to the issue
- [X] Labels are assigned to the issue

rsh52:
# Description
This PR addresses #31. It was found that for checkbox variable types, raw values have `-`s converted to `_`s once exported from REDCap. To correctly output checkbox variables, `checkbox_appender` needs to enforce this change similarly to how it enforces letters to lowercase with `tolower`. This change only affects raw data assignment.

I do not believe we need a new test for this, instead I added a value to our sample database that contains `-`s (shown in screenshot). A successful `read_redcap_tidy` call indicates it's working.

**Supporting Screenshots**

REDCap UI:
image

Output:
image

# Proposed Changes 

- Add `str_replace_all` to `checkbox_appender` and replace `-`s with `_`s

### Issue Addressed
**Closes #31**

### PR Checklist

- [NA] New/revised functions have associated tests

# Code Review

### Code Review Checklist


skadauke:
Is this a REDCapTidieR issue or a REDCapR issue??

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

rsh52:
> Is this a REDCapTidieR issue or a REDCapR issue?? Get Outlook for iOS<https://aka.ms/o0ukef>

It's a `REDCapTidieR` issue, if you export the database right from the REDCap front end export tool the checkbox variables have `-`s converted to `_`s in the variable names. This is similar to how we're enforcing `tolower` which is also a REDCap default behavior. It's an issue for us since we're using the metadata to inform the database name outputs.

pwildenhain:
With this change, `redcap_read_tidy()` now works for me 🥳

skadauke:
Before we merge this - is that how things _should_ work? Wouldn't it be better to do it the other way round? 

Part of the scope of REDCapTidieR is to make things less confusing for the analyst. I think it's confusing if a raw category of `-99` gets turned into `checkbox____99`. Wouldn't `checkbox___-99` be better? 

@pwildenhain , @rsh52 thoughts?

pwildenhain:
Agree that something like `my_checkbox___-99` is a less confusing column name

Two thoughts: 
1. By changing the export field name you would contradict the API (and `REDCapR` by association). Not necessarily bad, but something I would be hesitant to do
2. If you leave the minus sign in there, then the analyst has to use backticks whenever they reference the variable 

```r
df |>
  select(record_id, `my_checkbox___-99`
```

skadauke:
Let's make it less confusing then. The primary intent behind REDCapTidieR is to reduce the cognitive load of an analyst as they're doing exploratory data analysis. That means that we're modifying the data in many ways from the way the API returns it. 

I think the backticks thing is fine. 

Finally, I would as a guideline use human-readable raw values for checkboxes in REDCap.

For example, instead of defining a hypothetical `gender` checkbox like so:

```
1, Male gender
2, Female gender
-99, Unknown gender
```

you should use:

```
male, Male gender
female, Female gender
unknown, Unknown gender
```

That way what you'll get from `REDCapR` or `REDCapTidieR` is `gender___unknown` instead of `gender____99` or `gender___-99`. And you also avoid the whole `-` vs `_` trap.

This may seem like an extra hassle but it's not so bad if you consider that we're only talking about multi-checkbox fields here. Not necessary to do this for radio buttons or dropdowns.

(PS. I know `gender` example doesn't make any sense as a multi field but whatev)

pwildenhain:
Ok got it -- so what does the fix look like in that case? As of right now, `redcap_read_tidy()` throws an error for a checkbox field with `-99` as a value

skadauke:
> Ok got it -- so what does the fix look like in that case? As of right now, `redcap_read_tidy()` throws an error for a checkbox field with `-99` as a value

The function will continue to throw errors until we've fixed the issue and merged the PR. :)

We'll need to add logic to handle reversing the `-` to `_` replacement done by the REDCap API. This isn't as simple as replacing all `_` with `-` in `db_data`. I think to do this correctly, we'll have to match the fields in `db_data` with the corresponding field in `db_metadata` and then replace the name of the field in `db_data` with what's in the `db_metadata`.

skadauke (approved):
This looks good to me!

rsh52:
Awesome! Closing and merging along with #33. @pwildenhain You should be able to use this for your database now.

PR #41, opened 2022-09-29
block/sparse matrix terminology

wibeasley:
I doubt many people would notice or care, but the mashup table (ie, the thing REDCapTidieR addresses) is technically a [block matrix](https://en.wikipedia.org/wiki/Block_matrix).  

A [sparse matrix](https://en.wikipedia.org/wiki/Sparse_matrix) is something else.  Actually it is spiritually equivalent to REDCap's underlying EAV/`redcap_data` table.

To make things a little more confusing (blame the linear algebra people, not me), a block matrix can be called "sparse" if it has a lot of missing/zero values.  There are a few occurrences of "sparse" that I didn't replace because it made sense for describing an empty-ish block matrix.

---

I've never used renv, but I've always like the concept.  Does it normally take ~20 min to set up for each package you develop?


skadauke (approved):

skadauke:
Thanks for this clarification! I'm not sure that it should take 20 min to get `renv` going... @rsh52?

skadauke:
So, the tests failed because `Assertion on 'redcap_uri' failed: Must comply to pattern '^.{1,}$'.`. Assuming this is because  `REDCAP_URI` env var wasn't set for some reason. I'm not sure why. I noticed that our secrets are in the Org context: 

image

Maybe because Will isn't in our Org, his CI runs don't have access to Org secrets? Might need to move them to a repo context. @rsh52, can you take a look?

wibeasley:
> Maybe because Will isn't in our Org, his CI runs don't have access to Org secrets? 

As I was submitting it, I wondered if there's extra trust required.  Otherwise I could slip in some code that sends the password/token/secret to some webserver or maybe just dump it in a GitHub Actions artifact

rsh52:
Looks like we need to set up `renv` in our actions workflows to cache packages and that'll solve the time issue: https://rstudio.github.io/renv/articles/ci.html#github-actions


So far as the secrets failure, I don't think it's related to their availability (they're open to all public repos in our org) but it might be due to the requirement of a first approval since this is a forked PR:

image

skadauke:
Apparently this is intended behavior to prevent what's called a **Pwn Request**. Not a day goes by without learning something new.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

There are some workarounds described in the article above. The one I think might work best for us is described here:

image

rsh52:
@skadauke sounds good to me, I like this approach a lot. Sets up a nice checkpoint that maintains control and works well inside of GitHub. Can implement Monday!

skadauke (approved):

PR #45, opened 2022-10-07
Implement httptest in tests and vignettes

### LINKED ISSUE #44: [FEATURE] Mock API calls with httptest so tests and vignettes can run on CRAN
# Feature Request Description
Tests of `read_redcap_tidy()` and the bulk of our vignette currently don't run on CRAN to avoid making live API calls during testing and exposing credentials. `httptest` has the ability to create mock API responses that can be fed back to `read_redcap_tidy()` during testing/vignette rendering to simulate real API responses.

This needs to be done in such a way that REDCap tokens (and ideally the CHOP REDCap uri) are not stored publicly or needed to run tests/vignettes.

# Proposed Solution
1. Add a script in `/tests` to create mocks for everything in `test-read_redcap_tidy.R` and store as test fixtures
    - Create [custom redactor](https://enpiar.com/r/httptest/articles/redacting.html#writing-custom-redacting-functions) to replace tokens and uri with fakes that can be referenced in tests and vignettes
2. Follow [these instructions](https://enpiar.com/r/httptest/articles/vignettes.html) to set up vignettes to use mocks

### Checklist

- [x] The issue is atomic

ezraporter:
# Description
This PR adds the infrastructure to run tests and vignettes using `read_redcap_tidy()` (or any API calling function) without actually calling the API and without storing credentials or the CHOP redcap uri.

The process is facilitated by a custom "redactor" function (see `inst/httptest/redact.R`) that gets called when mocks are being made. The redactor swaps out real credentials for fake credentials and then saves the mock as if the request had been made with the provided fake credentials. The resulting workflow is:

1. Any developer with valid credentials in their `.Renviron` runs vignettes or creates mocks for tests by running `tests/testthat/fixtures/create-httptest-mock.R`
2. Anyone (including CRAN) can run api calls we have mocks for using the fake credentials if:
    - the call is wrapped with `with_mock_api()` or
    - the call occurs in a vignette that includes `httptest::start_vignette()`
3. If new mocks need to be added:
    - For tests: add to `create-httptest-mock.R` and run it to regenerate
    - For vignettes: delete the `vignettes/name-of-vignette` directory and knit

# Proposed Changes 

- Add custom redactor in `inst/httptest/redact.R`
- Create testing mocks in `testthat/fixtures` along with the script to regenerate them
- Wrap all api calls in tests with `with_mock_api()`
- Add `start_vignette()` to vignettes and run them to create mocks
- Drop skip on CRAN logic since we should be able to have this run now!

# Additional info

This PR looks more bloated than it is because all the mocks need to be committed and github wasn't very friendly about `with_mock_api()` getting added to test files.

### Issue Addressed
Closes #44

### PR Checklist

# Code Review

### Code Review Checklist


vignettes/REDCapTidieR.Rmd
```diff
@@ -10,21 +10,37 @@ vignette: >
...
-redcap_uri <- Sys.getenv("REDCAP_URI")
+# If we're creating mocks we need to use the real tokens. Otherwise, use fakes.
+
+# Check whether we're mocking
+# start_vignette() starts mocking if the mock directory doesn't exist
+# Checking this way is slightly hacky but httptest doesn't seem to expose
+# anything that would allow an "official" check
+mocking <- !dir.exists("REDCapTidieR")
+
+if (mocking) {
```
skadauke, line 37: I'm not sure this works. What if you've added a new API call to the Rmd? In this case you'd have the fake creds set but nothing cached. 

I believe what you want to do here is use the env vars if they are available.

vignettes/REDCapTidieR.Rmd
```diff
@@ -52,6 +68,10 @@ Here is a screenshot of the REDCap Status Dashboard of this database. Note that
 
 Great! Now let's import the superheroes data into R. We can use `REDCapR::redcap_read_oneshot()` which returns a list with and element named `data` that contains all of the data as a data frame. We turned this data frame into a tibble for better readability:
 
+```r
+token <- "123456789ABCDEF123456789ABCDEF04"
```
skadauke, line 72: I think we set this already above? Remove this chunk?

vignettes/longitudinal_redcap.Rmd
```diff
@@ -8,19 +8,32 @@ vignette: >
...
+# TLDR: Don't move this library call
 library(REDCapTidieR)
 
-token <- Sys.getenv("REDCAPTIDIER_LONGITUDINAL_API")
-redcap_uri <- Sys.getenv("REDCAP_URI")
+# If we're creating mocks we need to use the real tokens. Otherwise, use fakes.
+
+# Check whether we're mocking
+mocking <- !dir.exists("longitudinal_redcap")
+
```
skadauke, line 29: See comment for the other vignette above

vignettes/longitudinal_redcap.Rmd
```diff
@@ -46,6 +59,11 @@ The `read_redcap_tidy` function returns a tibble in which each row represents a
 
 Let's use a sample longitudinal REDCap database containing repeat instruments, longitudinal events, and multiple event arms:
 
+```r
+token <- "123456789ABCDEF123456789ABCDEF02"
```
skadauke, line 63: See question for the other vignette above

skadauke (changes requested):
This is an amazing start! However I do have a few comments/questions that I think we need to address.

1. For some reason, the GHA checks didn't run. I think it's because we futzed around with the workflow file. We need to fix and make sure checks pass on GHA.
2.  I have a couple of workflow questions since we're making things a bit more complicated with essentially creating a cache that we will need to manually refresh, and so we need to decide when we want to do that and set ourselves specific reminders in the correct places (no good deed goes unpunished eh?) - ideas might include our PR checklist and for `devtools::release()` we can add specific `release_questions()`, such as:

vignettes/REDCapTidieR.Rmd
ezraporter: I think I follow but the way `start_vignette()` works means you will either be creating new mocks for the entire vignette or using existing mocks for the entire vignette. The [recommended workflow](https://enpiar.com/r/httptest/articles/vignettes.html) for adding a new mock like the case you described is to delete the directory with mocks for the vignette and then knit it to regenerate all of the mocks.

If we supplied env vars when the vignette isn't creating mocks it would fail because it would be looking for mocks with the fake credentials.

ezraporter:
Thanks! Looking into the GHA checks now. I'll learn a bit more about `release_questions()` and think about the workflow. Totally agree that we need to add structure for updating mocks along with the additional complexity.

vignettes/REDCapTidieR.Rmd
ezraporter: Ah interesting. I like that approach. Let me see if I can get it to work

vignettes/REDCapTidieR.Rmd
skadauke: Ah - gotcha! 🤦‍♂️

skadauke:
> 2 GHA questions:
> 
> 1. The reason checks were skipped initially is because the `safe to test` label was missing. Is our intended workflow to add that to all PRs (including internal ones) or should we look into a way to require that **only** when the PR is coming from a fork? (Haven't yet looked into whether that second option is possible/easy to implement)

I think if your most recent change set makes it GitHub secrets unnecessary then we should probably return to the previous workflow?

> 2. Our code coverage check is newly failing with `covr` issuing `Error: Please install xml2 package`. I'm not entirely clear on what changed. Looking at the logs it appears that we ran the job that failed with an updated version of `curl` relative to the last successful run:
> 
> [This run](https://github.com/CHOP-CGTDataOps/REDCapTidieR/actions/runs/3221475334/jobs/5269666582): `curl 4.3.3` [Last run](https://github.com/CHOP-CGTDataOps/REDCapTidieR/actions/runs/3190644226/jobs/5206024553): `curl 4.3.2`
> 
> Maybe me and Rich can look at it together when he gets back.

I think you could try editing `.github/workflows/test-coverage.yaml` to install the `xml2` package? Seems like the `setup-r-dependencies@v2` workflow allows specifying extra packages:

```
      - uses: r-lib/actions/setup-r-dependencies@v2
        with:
          extra-packages: any::covr
          needs: coverage
```

ezraporter:
> I think you could try editing .github/workflows/test-coverage.yaml to install the xml2 package? Seems like the setup-r-dependencies@v2 workflow allows specifying extra packages

Do I make updates to workflows directly on `main`? Seems like the change I made on this branch isn't getting picked up by GHA

skadauke:
It looks like you re-ran the workflow? I think if you trigger a re-run the same workflow file gets run. Not sure what's the best way to manually trigger a rerun of an updated workflow file.

skadauke:
Looks like for now, removing and adding back `safe to test` did the trick

ezraporter:
I think I sorted this out. There were multiple problems.

1. Our `pull_request_target` workflow wasn't quite setup correctly. The key is that when you use `pull_request_target` the branch being operated on is the **target branch** (in this case `main`) by default. If you want to test code in the feature branch you need to explicitly check it out. [Here's](https://github.com/ezraporter/gha-test) a little reprex repo I used to convince myself of that.

2. Because of 1 we actually did have to make changes on `main` for them to get picked up correctly. This is because `pull_request_target` causes the GHA to use the version of the workflow on the target branch. I mocked that up [here](https://github.com/ezraporter/gha-test/commit/d04542b2692b0cd8020625be638afd51648ec258) and you can see what happened [here](https://github.com/ezraporter/gha-test/actions/runs/3227587967/jobs/5282589296).

Rather than fixing the `pull_request_target` workflow I just reverted `pull_request` and removed the credentials since we potentially don't need them anyways.

ezraporter:
A couple changes to wrap things up:

1. Got @skadauke's cache invalidation suggestion working for vignettes. I was actually able to save us a little duplication by putting the set up code in `inst/httptest/start-vignette.R` as suggested [here](https://enpiar.com/r/httptest/articles/vignettes.html#advanced-topics) at the very end of the page.

2. Added a csv with our fake credentials in `inst/misc` and created a helper function to access them

ezraporter:
Also I noticed our GHA workflows are throwing some new warnings related to [this announcement from GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). Looks like there's already an issue on r-lib/actions: https://github.com/r-lib/actions/issues/627

skadauke (approved):
This is great! I think we should also delete the secrets from our GitHub org. I don't think I have access to them - @rsh52, can you take care of that?

rsh52:
> A couple changes to wrap things up:
> 
> 1. Got @skadauke's cache invalidation suggestion working for vignettes. I was actually able to save us a little duplication by putting the set up code in `inst/httptest/start-vignette.R` as suggested [here](https://enpiar.com/r/httptest/articles/vignettes.html#advanced-topics) at the very end of the page. Everything in `start-vignette.R` runs when you call `start_vignette()`.
> 2. Added a csv with our fake credentials in `inst/misc` and created a helper function to access them

Excellent, love this and leaning on `inst/` to carry the weight of the `httptest` custom functions. Good job with reducing some of the other helper functions as well.

> Also I noticed our GHA workflows are throwing some new warnings related to [this announcement from GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). Looks like there's already an issue on r-lib/actions: [r-lib/actions#627](https://github.com/r-lib/actions/issues/627)

Might be worth putting as a new Issue, seems like this is extremely recent but may be a necessary change or note in the next CRAN release.

rsh52 (approved):
See comments above, all set!

ezraporter:
> Might be worth putting as a new Issue, seems like this is extremely recent but may be a necessary change or note in the next CRAN release.

Makes sense! I'll make an issue for it

rsh52:
> This is great! I think we should also delete the secrets from our GitHub org. I don't think I have access to them - @rsh52, can you take care of that?

Done!

PR #51, opened 2022-10-18
Fix bug causing NAs when first form in project is not requested

### LINKED ISSUE #47: [BUG] read_redcap_tidy returns incorrect results when forms are specified that don't include first instrument in the project
# Expected Behavior
Regardless of which instruments are requested, `read_redcap_tidy()` should return an object with each instrument-tibble containing:
- identifiers (`record_id`, `redcap_repeat_instance`, `redcap_event`, `redcap_arm`)
- one row per unique combination of identifiers in the project data

# Current Behavior
When the `forms` parameter is used and the vector of instruments doesn't include the instrument containing `record_id`, identifiers aren't included in the output.

When this occurs for longitudinal projects the output additionally contains extra rows with `NA` values.

## How to Reproduce the Bug:
Non-longitudinal:
```r
token <- Sys.getenv("REDCAPTIDIER_CLASSIC_API")

read_redcap_tidy(redcap_uri, token, forms = "repeated")
```

Longitudinal:
```r
token <- Sys.getenv("REDCAPTIDIER_LONGITUDINAL_API")

read_redcap_tidy(redcap_uri, token, forms = "repeated")
```

# Suggested Fix
This seems to be occurring because the REDCap API only returns identifiers if the instrument containing `record_id` is included in the `forms` argument. Augmenting the API call so that the form with `record_id` is always requested under the hood should resolve the issue.

### Checklist

- [x] The issue is atomic

ezraporter:
# Description
This PR addresses the issue with incorrect data being returned when a user passes a `forms` argument but doesn't include the form containing identifiers. The fix is to ensure that we're always including the first form in the project in our call to `redcap_read_oneshot()` when a user passes a `forms` argument. We then get some extra non-identifier fields from that form and need to filter those out of the result.

We were previously not catching this issue in our tests because we were:
1. Only testing the `forms` parameter for the classic redcap
2. Using a test case where the `forms` argument was also the first form in the project

To increase our coverage I added a test for the longitudinal redcap and updated this and the old test to use a form other than the first one in the project.

# Proposed Changes 

- Add logic to insert the first form in the metadata into our `forms` argument to `redcap_read_oneshot()`
- Add logic to determine which fields need to be kept in the results of `redcap_read_oneshot()`
    - This is stored in a helper function, `get_output_fields()`

# Additional Notes

This change also required me to update the mocks since we tweaked the test cases

### Issue Addressed
Closes #47

### PR Checklist

# Code Review

### Code Review Checklist


rsh52 (approved):
Great work 🎉

Issue #48, opened 2022-10-17
[BUG] Improve REDCapTidieR Execution Times and Benchmarking


rsh52:
# Expected Behavior
`REDCapTidieR` should run relatively quickly (<10s) on most manageable databases. 

# Current Behavior
Currently `REDCapTidieR` can take long stretches of time that are not related to API calls. Using `profvis` it was found that the main culprit here is the `check_repeat_and_nonrepeat` function that gets looped through inside of the `clean_*` functions. This is a nested `for` loop that looks across rows and columns of a database to check whether a project is repeating or nonrepeating.

The complicating factor here is the allowed REDCap behavior to design an instrument that can be _both_ repeating **and** nonrepeating. This is a behavior we do not want to support, but wish to loudly fail on. If we can fix the below chunk for this check, it should greatly improve execution time:

https://github.com/CHOP-CGTDataOps/REDCapTidieR/blob/1c70017dff24b1600bfe18d79695af3ff148ae5e/R/checks.R#L67-L89

There may be some other areas that can marginally improve execution time, but this seems to be the biggest driving factor.

**Screenshots**
`profvis` outputs shown below:
![image](https://user-images.githubusercontent.com/38384694/196259221-37b35c15-0cb9-4ea9-8d16-1f260b2b2ea8.png)
![image](https://user-images.githubusercontent.com/38384694/196259278-770fe2d6-c69c-446d-bb6d-bd7398a6b37a.png)

### Checklist
Before submitting this issue, please check and verify below that the submission meets the below criteria:

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

ezraporter:
Some updates after diving deeper in the `profvis` output.

1. I think we were being misled by `profvis` a bit for reasons I'm not sure of. This is the current output from `profvis` in table form:
![profvis-tabular](https://user-images.githubusercontent.com/60618324/196793679-19caeb29-5e1c-43ed-9db2-e8bf250d6a32.png)

For some reason it's only attaching a small subset of the operations that are happening to `read_redcap_tidy`. The other portions are **definitely** things `read_redcap_tidy` is doing but `profvis` makes it hard to figure out what's happening where.

[...]

Issue #16, opened 2022-08-14
[FEATURE] Augment the supertibble with metadata 


skadauke:
# Feature Request Description

The supertibble should make it easy for the analyst to discover table-specific metadata. 

# Proposed Solution

Add the following columns to the supertibble returned by `read_redcap_tidy()`:

- `redcap_form_label`: this should be the instrument label returned by `redcap_instruments` 
- `redcap_metadata`: This should be a list column containing data from `db_metadata` specific to the form (i.e. filtered on `form_name`). This will include non-data fields, which I think is OK. I suggest the following modifications from the raw REDCapR output: (1) remove `form_name` as it's redundant; (2) Turn `select_choices_or_calculations` into a list column with tibbles generated by `parse_labels`; (3) Add a list column `events` that contains a tibble of events which which the instrument is associated - skip if not longitudinal. If there are arms in the project, this tibble should have a column "arm" to indicate arms. (4) Reorder columns so the most useful ones are on the left: `field_name`, `field_label`, `field_type` then `everything()` else
- `rows` showing the number of rows of the tibble
- `columns` showing the number of cols of the tibble
- `NA` percentage
- object size

We might want to add more informational columns. For inspiration, see here: https://cghlewis.github.io/codebook-pkg-comparison/
Note, see: https://github.com/skadauke/dataMeta

## Potential Related Enhancement

Use the [`labelled`](https://larmarange.github.io/labelled/) package to add labels from the metadata to fields in the super-tibble as suggest on the CHOP R User Group slack

### Checklist

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

ezraporter:
I've implemented the basic functionality + tests [here](https://github.com/CHOP-CGTDataOps/REDCapTidieR/tree/add-metadata). Two things:

1. I'm waiting on adding the `events` column until #56 is merged. The refactor of `link_arms()` is just involved enough that it doesn't make sense to recreate it independently on this branch. Merging that and rebasing this branch will be much cleaner.

2. In thinking about some of the bonus elements like number of rows, number of columns, NA%, object size I wonder if we should relegate those things to a `summary()` method. The functionality would be:

```r
redcap_data <- read_redcap_tidy(uri, token)

summary(redcap_data)

#> form  field nrow ncol na_percent ...
#> abc   f1    10   2    0
[...]

PR #56, opened 2022-10-21
Remove looping over longitudinal events in distill_*_table_long

ezraporter:
# Description
This PR addresses the first and largest bottleneck described in https://github.com/CHOP-CGTDataOps/REDCapTidieR/issues/48#issuecomment-1285717234. Currently for each form we are looping over longitudinal events and sequentially filtering the data to remove events that aren't mapped to the form. A faster solution is to aggregate all the event mappings for the form upfront and filter to only the events mapped to the form. (Would like to get Rich's thoughts to make sure I'm understanding the subtleties of longitudinal redcaps!)

I also modified the `link_arms()` function to return a `tibble` in a more standard format. That has two benefits:
1. Makes it slightly easier to grab all the events associated with a form
2. Returns form-event mappings in a friendlier format that we can use as metadata for #16 

# Proposed Changes 

- Refactor `link_arms()` to return a more standard event-form `tibble`
- Refactor `distill_*_table_long()` functions to avoid looping over longitudinal events

# Additional Info

1. Tests are currently failing because we need to rerun `inst/testdata/create_test_data.R`. I'm missing 2 redcap API keys to do that:
```r
classic_norepeat_token <- Sys.getenv("REDCAPTIDIER_CLASSIC_NOREPEAT_API")
longitudinal_norepeat_token <- Sys.getenv("REDCAPTIDIER_LONGITUDINAL_NOREPEAT_API")
```

2. I had to make similar changes to both `distill_repeat_table_long()` and `distill_nonrepeat_table_long()`. Should we consider combining them into a single function? It looks like there's non-trivial code duplication.

3. This shaved about 4 seconds off our run time with the prodigy db:

```r
devtools::load_all()
library(microbenchmark)
library(profvis)

redcap_token <- Sys.getenv("PRODIGY_REDCAP_API")

microbenchmark(read_redcap_tidy(redcap_uri, redcap_token), times = 10)

#> Unit: seconds
#>                                       expr      min       lq     mean   median       uq      max neval
#> read_redcap_tidy(redcap_uri, redcap_token) 9.812751 10.34588 10.61901 10.55411 11.03919 11.44642    10
```

```r
profvis::profvis(read_redcap_tidy(redcap_uri, redcap_token), simplify = FALSE)
```
Screen Shot 2022-10-21 at 12 51 03 PM

We're actually spending less time here than in the checkbox updating now.

### Issue Addressed
Relates to #48 

### PR Checklist

# Code Review

### Code Review Checklist


skadauke:
This is great! Looks like we already decreased runtime by half??

I had actually discussed with @rsh52 previously to refactor `link_arms()` to return a tibble. At the time we decided to leave it alone because it was working. But I agree, this is a good opportunity to deal with some tech debt. 

I also agree with the idea of combining `distill_repeat_table_long` and `distill_nonrepeat_table_long` if it's easy to do. Would probably prioritize this lower than any of the outstanding bug reports and feature requests however. I would say, if you think you can tackle this as part of this PR with less than 3 hours of work, go for it.

@rsh52, can you provide all REDCap API keys to Ezra so we can get the tests to pass?

rsh52 (changes requested):
> @rsh52, can you provide all REDCap API keys to Ezra so we can get the tests to pass?

@ezraporter I looked through all of the REDCap databases, you currently have API privilege access to all of them but it doesn't seem like a token has been generated for "**[REDCapTidier Test - Classic No Repeat](https://redcap.chop.edu/redcap_v12.5.8/UserRights/index.php?pid=53972)**" and "**[REDCapTidieR Test - Longitudinal No Repeats](https://redcap.chop.edu/redcap_v12.5.8/UserRights/index.php?pid=53974)**". I can give you mine if necessary, but prefer if we keep API keys to our own management. Let me know if this still doesn't work and I'll cease my grumbling.

> I had to make similar changes to both distill_repeat_table_long() and distill_nonrepeat_table_long(). Should we consider combining them into a single function? It looks like there's non-trivial code duplication.

If we can do it, let's! Good time to consolidate. My guess is we totally can, but at the time of writing these functions we wanted to keep their responsibilities separate until we had a clearer picture of what we wanted them to do.

Skimmed through the changes a bit this morning and so far really liking the improvements, much cleaner and easier to read. Also realizing a test was never written for `link_arms`, should write one especially now that it has a more manageable output. We should write that before merging, then when all the checks pass this should be good to go! Fantastic work Ezra 🎉

ezraporter:
> If we can do it, let's! Good time to consolidate. My guess is we totally can, but at the time of writing these functions we wanted to keep their responsibilities separate until we had a clearer picture of what we wanted them to do.

Actually any objections to making this a separate issue and handling it in a new branch? I'm hesitant to make the diff on this one any larger.

rsh52:
> > If we can do it, let's! Good time to consolidate. My guess is we totally can, but at the time of writing these functions we wanted to keep their responsibilities separate until we had a clearer picture of what we wanted them to do.
> 
> Actually any objections to making this a separate issue and handling it in a new branch? I'm hesitant to make the diff on this one any larger.

Yes, definitely. Better to get these changes finalized and implemented then move on to the next ones.

ezraporter:
Should be good for review once the checks finish. I had to move a couple things around to add the `link_arms()` test since it's an API calling function so should use `httptest`.

skadauke (approved):
LGTM!

rsh52 (approved):
Ditto, thank you for writing up the test and updating the mocks. Good to merge!

Issue #46, opened 2022-10-12
Address deprecation warnings in GitHub Actions workflows


ezraporter:
# Description

Our Actions workflows began issuing warnings related to [this update](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) from GitHub.

![Screen Shot 2022-10-12 at 3 39 01 PM](https://user-images.githubusercontent.com/60618324/195432922-c5e8b6ee-a3ec-4642-adee-296ca2cf9433.png)

There's an open issue on r-lib/actions https://github.com/r-lib/actions/issues/627

### Checklist

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [ ] Developers are assigned to the issue
- [ ] Labels are assigned to the issue

skadauke:
Not sure we need to/should do anything about this? Once `r-lib/actions` is fixed, this warning should automagically go away for us, right?

rsh52:
You're probably right, I was wondering if the fix would involve any changes to the actions callout (like `@v2` to `@v3`), but that likely wouldn't happen. Figured that since we have a warning returned from our Actions runs it would be worth documenting the reason why and then we can close this once the `r-lib` folks resolve it.

skadauke:
Totally agree!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Rich Hanna ***@***.***>
Sent: Thursday, October 13, 2022 8:15:02 AM
To: CHOP-CGTDataOps/REDCapTidieR ***@***.***>
Cc: Stephan Kadauke ***@***.***>; Comment ***@***.***>
Subject: [External]Re: [CHOP-CGTDataOps/REDCapTidieR] Address deprecation warnings in GitHub Actions workflows (Issue #46)


You're probably right, I was wondering if the fix would involve any changes to the actions callout (like @v2 to @v3), but that likely wouldn't happen. Figured that since we have a warning returned from our Actions runs it would be worth documenting the reason why and then we can close this once the r-lib folks resolve it.

—
[...]

PR #58, opened 2022-10-26
Supplement supertibble with metadata

### LINKED ISSUE #16: [FEATURE] Augment the supertibble with metadata

ezraporter:
# Description
This PR supplements the supertibble resulting from `read_redcap_tidy()` with additional metadata fields. See the updated documentation for `read_redcap_tidy()` to see what we're adding.

# Proposed Changes 

- Add an `include_metadata` parameter to `read_redcap_tidy()` to control whether metadata fields are added
- Add 2 new functions called at the end of `read_redcap_tidy()` to add the metadata
    - `add_metadata()` adds metadata relevant for both longitudinal and non-longitudinal projects
    - `add_event_mappings()` is called only if the project is longitudinal and adds the event/arm information for each instrument
- Add additional processing to the metadata in `read_redcap_tidy()` to dissociate the project record_id from a specific form
    - This is a small technical change to make sure that all of the metadata for the record_id field is captured if the user uses the `forms` parameter but doesn't include the first form in the project.

### Issue Addressed
Closes #16 

### PR Checklist

# Code Review

### Code Review Checklist


rsh52 (commented):
I think this all looks very good and followable, the mock updates could just be because the API changed.

Once we get actions/R CMD Check to pass I think this is ok to merge. It looks like the `pak` package was just added to r-lib `setup-r-dependencies` [7 days ago](https://github.com/r-lib/actions/tree/v2-branch/setup-r-dependencies) and is just an alternate way to install R packages. [This issue](https://github.com/r-lib/actions/issues/652) brings this up as well and the author has a suggested solution from an hour ago.

ezraporter:
> [This issue](https://github.com/r-lib/actions/issues/652) brings this up as well and the author has a suggested solution from an hour ago.

I'm actually not sure this solution helps us because we're already doing the suggested solution of having `install-r: true`. It looks like `true` is the default for [`setup-r`](https://github.com/r-lib/actions/tree/v2/setup-r#inputs) and we're not setting it currently:

https://github.com/CHOP-CGTDataOps/REDCapTidieR/blob/70a85588b27cd7186fd2dfb521b1820dfcd358ed/.github/workflows/R-CMD-check.yaml#L26-L28

The issue may be with `renv`. I made a little example [here](https://github.com/ezraporter/linrtest/tree/branch). [This run](https://github.com/ezraporter/linrtest/actions/runs/3337385681/jobs/5523661053) fails with `renv` activated in the project showing the same `pak` error. But [this run](https://github.com/ezraporter/linrtest/actions/runs/3337406762/jobs/5523708174) passes once I ran `renv::deactivate` on the project.

Only problem is I'm not sure how to resolve the `renv` issue in our package since shutting it off isn't what we want to do.

ezraporter:
Well it's a better solution than having it in our lockfile. Once the last checks pass on that branch I'll review then once we merge it I can update from `main` to pull the updated workflow into this branch.

tests/testthat/test-read_redcap_tidy.R
```diff
@@ -254,3 +254,61 @@ test_that("get_fields_to_drop handles checkboxes", {
...
+  # metadata contains all expected fields in data
+
+  ## fields we don't expect
+  exclude_fields <- c(
+    "redcap_repeat_instance", "redcap_event",
+    "redcap_arm", "form_status_complete"
+  )
+
+  fields_in_metadata <- out$redcap_metadata %>%
+    map("field_name") %>%
```
skadauke, line 296: This (ll. 295-301) is hard to understand. Can you rewrite w/o `map` functions?

skadauke (changes requested):

R/read_redcap_tidy.R
```diff
@@ -55,14 +77,19 @@ read_redcap_tidy <- function(redcap_uri,
                              token,
                              raw_or_label = "label",
                              forms = NULL,
-                             suppress_messages = TRUE) {
+                             suppress_messages = TRUE,
+                             include_metadata = TRUE) {
```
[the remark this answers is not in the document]
ezraporter: This one got resolved in the last commit: https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/58/commits/65d532cba6b25b8f132df3dfd5afb0468f292075 ✅

R/read_redcap_tidy.R
```diff
@@ -158,6 +189,15 @@ read_redcap_tidy <- function(redcap_uri,
...
+  # Augment with metadata ----
+  if (include_metadata) {
+    out <- add_metadata(out, db_metadata, redcap_uri, token)
+
+    if (is_longitudinal) {
+      out <- add_event_mapping(out, linked_arms)
+    }
+  }
+
   out
```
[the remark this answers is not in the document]
ezraporter: We are but in kind of a roundabout way. I'll update to do it explicitly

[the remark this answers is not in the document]
ezraporter: This is now in the updated version. We also explicitly test for the ordering here:
https://github.com/CHOP-CGTDataOps/REDCapTidieR/blob/78ce9505122fe21325589e56c0fdd2236e352310/tests/testthat/test-read_redcap_tidy.R#L262-L269

tests/testthat/test-read_redcap_tidy.R
ezraporter: I'm not sure I can get away without mapping because this check tests that data fields and metadata fields match *within* each instrument (i.e. row of supertibble). I tried to clean things up and add clearer comments here: https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/58/commits/78ce9505122fe21325589e56c0fdd2236e352310. Let me know if that looks okay

skadauke: I think it looks good!

tests/testthat/test-read_redcap_tidy.R
```diff
@@ -254,3 +254,67 @@ test_that("get_fields_to_drop handles checkboxes", {
...
+    "data_na_pct"
+  )
+
+  # metadata fields exist and correctly ordered
+  expect_equal(expected_cols, names(out))
+
+  # metadata fields have the correct data types
+
+  ## redcap_metadata and redcap_events fields consist of tibbles
+  c(out$redcap_metadata, out$redcap_events) %>%
```
skadauke, line 274: I thought I left a comment but somehow it disappeared. Can we do this one without `map` functions? I'm thinking:

``` r
expect_s3_class(out$redcap_metadata, "tbl")
expect_s3_class(out$redcap_events, "tbl")
```

I know this is nitpicky but I really think that it's good to keep test code as easy to read as possible (especially for someone like myself who doesn't code every day!)

skadauke (changes requested):

tests/testthat/test-read_redcap_tidy.R
ezraporter: No worries! If want to do away with the `map` I think we'd need to just test the first element:
```r
expect_s3_class(out$redcap_metadata[[1]], "tbl")
expect_s3_class(out$redcap_events[[1]], "tbl")
```

`out$redcap_metadata` itself isn't a `tibble` it's a `list` so you get:
```r
expect_s3_class(out$redcap_metadata, "tbl")
#> Error: out$redcap_metadata is not an S3 object
```

Or if we're feeling really fiesty:
```r
expect_s3_class(sample(out$redcap_metadata, 1), "tbl")
expect_s3_class(sample(out$redcap_data, 1), "tbl")
```

to get a random element of out$redcap_* each time

rsh52: Unfortunately the most succinct way to make this test that I could think of is the way you wrote with `map_lgl`. The `sample` method shores up the test with randomness, but I don't really like that you technically can't say a test is the same from one run to the next. Doesn't seem kosher to me.

I think the best bet is to do indexing, and since the full length of what you're testing is (I believe) 6 tables, you could just write 6 expectation functions. It's not pretty, but we've done it in the past and it's readable.

skadauke: Ah, I see the difference now! And I'm just realizing for the first time that when you have a list column you can actually have different object types in different rows of a `tibble`! 😨 

So the `map` approach is more conservative from that point of view but I think checking the first element is good enough. I think the chance that someone will introduce a code change that would cause you to have a `tibble` in row 1 and some kind of other object in a different row is negligible.

The `sample` approach would introduce non-deterministic behavior into the test suite. There is a place for doing this kind of testing, but this is not that place :)

ezraporter: Switched things to just test the first element. I would be hard pressed to think of a case where we *actually* accidentally create a list column with mixed object types. Although yes technically possible. Testing the first element seems good enough.

rsh52:
@ezraporter Before merging this we should also include and check that this closes my recent comment in #46. From what I can see you are already updating all workflows checkouts to `@v3` (👏 ) which should handle the recent Node warning.

skadauke (approved):
LGTM!

Issue #70, opened 2022-11-09
[FEATURE] Improve `field_label` metadata for checkbox fields


ezraporter:
# Feature Request Description
Currently when checkbox fields are expanded in the metadata to have one row per checkbox option the field label from the checkbox field is carried over to all new rows. Field labels for each new row should be updated to include the option label in addition to the field label.

This addresses the conversation in https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/65#issuecomment-1306177174 

# Proposed Solution
Update `update_field_names()` to append field labels with option labels. See https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/65#issuecomment-1306177174 for example output.

### Checklist

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

skadauke:
This logic will need to be able to deal with embedded fields that are part of the choice label. E.g.:

field_name | field_label | select_choices_or_calculations
-- | -- | --
crs_anti_tcell_therapy_types | Anti-T Cell therapy given: | cyclophosphamide, Cyclophosphamide \| atg, ATG \| other, Other {crs_anti_tcell_therapy_types_specify}

The field label should be: 

`Anti-T Cell therapy given: Other` 

and not 

`Anti-T Cell therapy given: Other {crs_anti_tcell_therapy_types_specify}`. 

There may be variable amount of whitespace before the opening curly bracket. I don't know if there will ever be more than one embedded field but you never know. I'm thinking a regex that removes everything bounded by curlies and then strips whitespace from the right end of the string would do the trick.

PR #65, opened 2022-11-03
Implement labelling

ezraporter:
# Description
This PR implements the `make_labelled()` functionality described in #64. At this stage the core functionality with test cases 
 and reasonable error messages is complete. The remaining additions to complete the issue are:

- Implement custom formatter parameter in `make_labelled()`
- Add an exported default formatter
- Flesh out documentation
- Update vignette to show custom labelling + custom formatting functionality

# Proposed Changes 

Add exported `make_labelled()` function that:
- Applies predefined labels to columns of supertibble
- Applies labels to columns of elements of `supertibble$redcap_metadata` by replacing underscores with spaces and converting to title case
    - ex. `field_name` -> `"Field Name"`
    - I'm not set on this solution but it seemed to do the trick pretty well rather than predefining labels for all the metadata fields
-  Applies labels to columns of elements of `supertibble$redcap_data` based on the contents of `supertibble$redcap_metadata`

Input Checking:
- Input must be a `data.frame`/`tibble`
    - using `checkmate` for this
- Input must contain `redcap_data` and `redcap_metadata` columns
    - See error screenshot 1 
- Every element of `redcap_metadata` must have `field_name` and `field_label` columns
    - See error screenshot 2
- `labelled` must be installed
    - I'm using `rlang::check_installed()` for this. It does the check and also offers to install it for the user if it's missing. 

**Screenshots**
 Error 1:
Screen Shot 2022-11-03 at 1 22 23 PM

Error 2:
Screen Shot 2022-11-03 at 1 22 58 PM

<details>
  <summary>Code to produce errors</summary>

```r
devtools::load_all()

data <- tibble::tribble(
  ~ redcap_form_name, ~ redcap_data, ~ redcap_metadata,
  "form 1", tibble(x = letters[1:3]), tibble(field_name = "x", field_label = "X Label"),
  "form 2", tibble(y = letters[1:3]), tibble(field_name = "y", field_label = "Y Label")
)

data |>
  dplyr::select(-redcap_data) |>
  make_labelled()

data$redcap_metadata[[1]]$field_label <- NULL

make_labelled(data)
```
</details>

### Issue Addressed
Closes #64

### PR Checklist
Before submitting this PR, please check and verify below that the submission meets the below criteria:

- [x] New/revised functions have associated tests
- [x] New tests that make API calls use `httptest::with_mock_api` and any new mocks were added to `tests/testthat/fixtures/create_httptest_mocks.R`
- [x] New/revised functions use appropriate naming conventions
[...]

PR #73, opened 2022-11-10
Include checkbox field labels in metadata

ezraporter:
# Description
This PR updates our processing of the REDCap metadata to include checkbox option labels in the `field_label` column for each checkbox option row. The output differs slightly from what I described in #70:

Input:
|field_name|field_label|select_choices_or_calculations|
|----------------------|----------------------|-------------------------|
crs_anti_tcell_therapy_types | Anti-T Cell therapy given: | cyclophosphamide, Cyclophosphamide \| atg, ATG \| other, Other {crs_anti_tcell_therapy_types_specify}|

Output:
|field_name|field_label|updated_field_name|
|----------------------|----------------------|-------------------------|
crs_anti_tcell_therapy_types | Anti-T Cell therapy given: (Cyclophosphamide) | crs_anti_tcell_therapy_types___cyclophosphamide |
crs_anti_tcell_therapy_types | Anti-T Cell therapy given: (ATG) | crs_anti_tcell_therapy_types___atg |
crs_anti_tcell_therapy_types | Anti-T Cell therapy given: (Other) | crs_anti_tcell_therapy_types___other |

Two questions:
1. Not sure I like the carried over ":" in the middle of the field label but not sure how much additional formatting we should do by default. Should we do more? Currently we're just stripping field embedding + whitespace from the checkbox labels.
2. I had to make a decision on what happens if the checkbox field itself doesn't have a label. The current behavior is to print an explicit `"NA"` in the place of the field label before the checkbox option label. So:

|field_name|field_label|select_choices_or_calculations|
|---------|---------|----------|
my_field |  `NA` | 1, Option Label |

becomes

|field_name|field_label|updated_field_name|
|-------|-----------|--------------|
my_field |  NA (Option Label) |  my_field___1 |

Are there alternative suggestions?

# Proposed Changes

- Alter `update_field_names()` to edit `db_metadata$field_label` in place by appending checkbox option labels after the field label
    - This involved moving the `parse_labels()` call from `checkbox_appender()` into the body of `update_field_names()` so we can use both the labels and raw values. At that point enough of `checkbox_appender()` was whittled away that I just removed it.

### Issue Addressed
Addresses #70 (not closing until merged to main with #65)

### PR Checklist

# Code Review

### Code Review Checklist

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -173,73 +173,86 @@ parse_labels <- function(string, return_vector = FALSE) {
 #' Append REDCap checkbox variables with helpful labels
 #'
 #' @description
-#' Takes a \code{db_metadata$select_choices_or_calculations} field pre-filtered
-#' for checkbox \code{field_type}s and returns a vector of key names for
-#' appending to checkbox variables.
+#' Formats raw checkbox values by converting to lowercase and appending the
+#' name of the checkbox field from \code{field_name}
 #'
 #' @returns A vector.
 #'
 #' @param field_name The \code{db_metadata$field_name} to append onto the string
-#' @param string A \code{db_metadata$select_choices_or_calculations} field
-#' pre-filtered for checkbox \code{field_type}
+#' @param raw A vector of raw checkbox values produced by \code{parse_labels}
 #'
-#' @importFrom rlang .data
 #' @keywords internal
 
-checkbox_appender <- function(field_name, string) {
+checkbox_appender <- function(field_name, raw) {
   prefix <- paste0(field_name, "___")
-
-  out <- parse_labels(string)
-  out$raw <- tolower(out$raw)
-  # append each element of the split vector with the field_name prefix and then
-  # recombine
-  out <- paste0(prefix, out[[1]])
-
-  out
+  paste0(prefix, tolower(raw))
 }
 
 #' @title
 #' Update metadata field names for checkbox handling
 #'
 #' @description
-#' Takes a \code{db_metadata} object and appends a \code{field_name_updated}
-#' field to the end for checkbox variable handling.
+#' Takes a \code{db_metadata} object and:
+#' \itemize{
+#'   \item replaces checkbox field rows with a set of rows, one for each
+#'   checkbox option
+#'   \item appends a \code{field_name_updated} field to the end for checkbox
+#'   variable handling
+#'   \item updates \code{field_label} for any new checkbox rows to include the
+#'   specific option in parenthese
+#' }
 #'
-#' @returns Column \code{field_name_updated} appended to a given REDCap
-#' \code{db_metadata} object.
+#' @returns Column \code{db_metadata} with \code{field_name_updated} appended
+#' and \code{field_label} updated for new rows corresponding to checkbox options
 #'
 #' @param db_metadata The REDCap metadata output defined by
 #' \code{REDCapR::redcap_metadata_read()$data}
 #'
-#' @return Returns an updated REDCap metadata object with updated field names
-#'
 #' @importFrom tidyr unnest
+#' @importFrom dplyr %>% select mutate
+#' @importFrom tibble tibble
+#' @importFrom rlang .data
 #'
 #' @keywords internal
 
 update_field_names <- function(db_metadata) {
   out <- db_metadata %>%
     mutate(
-      field_name_updated = list(c())
+      updated_metadata = list(c())
     )
 
-  # Apply checkbox appender function to rows of field_type == "checkbox"
-  # Assign all field names, including expanded checkbox variables, to a list col
   for (i in seq_len(nrow(out))) {
     if (out$field_type[i] == "checkbox") {
-      out$field_name_updated[i] <- list(
-        checkbox_appender(field_name = out$field_name[i],
-                          string = out$select_choices_or_calculations[i])
+      # If checkbox field, parse labels fill updated_metadata with a tibble
+      # containing updated field names and labels
+      parsed_labs <- parse_labels(out$select_choices_or_calculations[i])
+
+      out$updated_metadata[i] <- list(
+        tibble(
+          field_name_updated = checkbox_appender(out$field_name[[i]], parsed_labs$raw),
+          # Append checkbox option to field label in parentheses
+          field_label_updated = paste0(out$field_label[[i]], " (", parsed_labs$label, ")")
+        )
       )
 
     } else {
-      out$field_name_updated[i] <- list(out$field_name[i])
+      # Otherwise carry through existing field name and label
+      out$updated_metadata[i] <- list(
+        tibble(
+          field_name_updated = out$field_name[[i]],
+          field_label_updated = out$field_label[[i]]
+        )
+      )
     }
   }
 
-  # Unnest and expand checkbox list elements
+  # Unnest and expand checkbox list elements + overwrite field_labels with
+  # updated field labels
+
   out %>%
-    unnest(cols = "field_name_updated")
+    unnest(cols = "updated_metadata") %>%
+    mutate(field_label = .data$field_label_updated) %>%
+    select(-"field_label_updated")
 }
 
 #' @title
--- a/R/utils.R
+++ b/R/utils.R
@@ -170,25 +170,6 @@ parse_labels <- function(string, return_vector = FALSE) {
 }
 
 #' @title
-#' Append REDCap checkbox variables with helpful labels
-#'
-#' @description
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -169,77 +169,94 @@ parse_labels <- function(string, return_vector = FALSE) {
...
+      clean_names <- paste(
+        out$field_name[[i]],
+        tolower(parsed_labs$raw),
+        sep = "___"
+      )
+
+      # Strip embedding logic and white space before appending to field labels
+      clean_labs <- parsed_labs$label %>%
+        # TODO: replace with formatting functions?
+        str_replace_all("\\{.*?\\}", "") %>%
```
rsh52, line 229: Did we confirm that this should be a default enforced option? Is there any reason we may want to have embedding/{} available, or added to the default `formatter` options? In my mind if we're not defaulting stripping HTML tags then we shouldn't default stripping embedded fields either.

rsh52 (commented):
Just one comment, otherwise I think this looks good and behaves well. Will definitely want the option to remove HTML once the `formatter` is ready. Great stuff!

R/utils.R
ezraporter: I think @skadauke and I hastily made that decision while meeting yesterday but I don't have strong opinions. If we chose not to edit the checkbox option labels on the way into the metadata the user would still be able to do things like strip html/embedding logic on the way into `labelled` labels when they call `make_labelled()`.

rsh52: I'm could see going either way since they serve different purposes, but I tend to lean toward showing the user everything and then letting them decide what they want to take away (or giving them the option to do so).

skadauke: To clarify - we're not talking about formatting labels for `labelled` functionality here but about changing the label of an option of a radio/checkbox field (although this also ultimately affects the tibble's variable labels). I believe that you _always_ want to remove field embedding code here.

skadauke:
> Two questions:
> 
> 1. Not sure I like the carried over ":" in the middle of the field label but not sure how much additional formatting we should do by default. Should we do more? Currently we're just stripping field embedding + whitespace from the checkbox labels.

Agree re: not carrying over `:`. I think what generally looks good is to have the field name followed by a colon and then the option label. It will look bad if there's extra whitespace and/or punctuation (`.`, `:`, `?`) here. I think, specifically for the purpose of generating checkbox field labels, we might want to strip extra whitespace and punctuation at the end of the field label, add a `:`, and then add the option label.

Not sure I'm a fan of enclosing the option label in parens. I think always having a colon between the field label and option label makes this unnecessary.

> 2. I had to make a decision on what happens if the checkbox field itself doesn't have a label. The current behavior is to print an explicit `"NA"` in the place of the field label before the checkbox option label. 

Great point. I think the best bet here will be to use information from the field name. We could remove underlines and make title case.  

So:

field_name	| field_label |	select_choices_or_calculations
----------|-----------------|------
my_field |	`NA` |	1, Option Label

field_name	 |field_label |	updated_field_name
----------|-----------------|------
my_field	| My Field: Option Label |	my_field___1

Do y'all think `make_labelled` should throw a warning if there are fields that don't have field labels?

R/utils.R
ezraporter: Correct. The question is what processing we should do to create the contents of `supertbl$redcap_metadata` before `labelled` gets involved

skadauke: It's a little bit confusing, but now that I'm thinking about it more there I believe that there are two different scenarios where field embedding code might pop up and they may need to be handled differently.

The first is a multi-checkbox such as in the example above. Here we'd want `Other {crs_anti_tcell_therapy_types_specify}` to be `Other`, and if this was the only scenario this could reasonably be handled with a formatter function.

The second scenario in which field embeddings can come up is in the context of a radio field (and possibly also dropdown but I'm not sure). In this case, `Other {something_specify}` doesn't affect the `labelled` variable field. But it _does_ affect the factor label of the variable which will be `Other {something_specify}` instead of `Other`.

ezraporter: Should we just give the user the option of applying a formatting function to all checkbox/radio/dropdown labels as an argument in `read_redcap_tidy()` or is that overly complex?

If yes we could centralize the logic in `parse_labels()` so that it would flow into both the metadata field labels (for checkboxes) and the factor levels (for everything else).

ezraporter:
> Agree re: not carrying over `:`. I think what generally looks good is to have the field name followed by a colon and then the option label. It will look bad if there's extra whitespace and/or punctuation (`.`, `:`, `?`) here. I think, specifically for the purpose of generating checkbox field labels, we might want to strip extra whitespace and punctuation at the end of the field label, add a `:`, and then add the option label.
> 
> Not sure I'm a fan of enclosing the option label in parens. I think always having a colon between the field label and option label makes this unnecessary.

I'm okay with this pending the discussion here re processing of options labels https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/73#discussion_r1021882354. Thoughts, @rsh52?

ezraporter:
> I think the best bet here will be to use information from the field name. We could remove underlines and make title case.  
> 
> So:
> 
> field_name	| field_label |	select_choices_or_calculations
> ----------|-----------------|------
> my_field |	`NA` |	1, Option Label
> 
> field_name	 |field_label |	updated_field_name
> ----------|-----------------|------
> my_field	| My Field: Option Label |	my_field___1

Using the variable name seems like a good idea. I wonder if doing the processing of removing underscores and converting to title case makes too many assumptions about the user's field names.

> Do y'all think make_labelled should throw a warning if there are fields that don't have field labels?

My gut reaction is that warning for this is too noisy. Mainly because:
1. Many REDCaps I've seen contain all kinds of hidden helper fields that you wouldn't want or need labels for
2. It's visually apparent in the viewer when labels are missing

I guess my combined personal answers to these questions are:
1. Use the variable name unedited for checkbox fields with no field label
    - Doing this will also keep it more visually clear when a label was missing but I'm also open to warning/messaging about this
    - i.e. `my_field is a checkbox field with no field label. Converting field labels to format "my_field: {checkbox option label}"`
2. In the general case of fields with no field labels I don't think we should warn

skadauke:
> > I think the best bet here will be to use information from the field name. We could remove underlines and make title case.
> > So:
> > 
> > 
> > 
> > field_name
> > field_label
> > select_choices_or_calculations
> > 
> > 
> > 
> > 
> > my_field
> > `NA`
> > 1, Option Label
> > 
> > 
> > 
> > 
> > 
> > 
> > field_name
> > field_label
> > updated_field_name
> > 
> > 
> > 
> > 
> > my_field
> > My Field: Option Label
> > my_field___1
> 
> Using the variable name seems like a good idea. I wonder if doing the processing of removing underscores and converting to title case makes too many assumptions about the user's field names.
> 
> > Do y'all think make_labelled should throw a warning if there are fields that don't have field labels?
> 
> My gut reaction is that warning for this is too noisy. Mainly because:
> 
> 1. Many REDCaps I've seen contain all kinds of hidden helper fields that you wouldn't want or need labels for
> 2. It's visually apparent in the viewer when labels are missing
> 
> I guess my combined personal answers to these questions are:
> 
> 1. Use the variable name unedited for checkbox fields with no field label
>    
>    * Doing this will also keep it more visually clear when a label was missing but I'm also open to warning/messaging about this
>    * i.e. `my_field is a checkbox field with no field label. Converting field labels to format "my_field: {checkbox option label}"`
> 2. In the general case of fields with no field labels I don't think we should warn

How about this: If field_label is `NA`, you don't get a variable label at all?

R/utils.R
ezraporter: I'm okay with this. I kind of agree with @rsh52's suggestion above https://github.com/CHOP-CGTDataOps/REDCapTidieR/pull/73#discussion_r1020283174 (which he may or may not still stand by) that if we're stripping embedding logic we also strip html.

I can't think of a reason a user would want raw html in their factor levels

rsh52: Yea, it's still my opinion that applying these changes should be done synonymously even if they serve different purposes. I could see the same person that says "I want to see the HTML" also saying "I want to see the field embedding". It may not be the most useful thing in 99% of cases but it's still a true representation of the output and may have its place somewhere. Some REDCap aficionado may be confused when things they expect to be there no longer are.

ezraporter:
We decided to:
1. Strip html and field embedding logic from all field labels and option labels by default in `read_redcap_tidy()`
2. Leave checkbox labels blank if `field_label` is `NA`
3. Not issue warnings for missing field labels

Will rebase on `main`, implement those changes, and open a new PR.

Issue #66, opened 2022-11-07
[BUG] read_redcap_tidy Output Order Doesn't Mirror Order in REDCap


rsh52:
# Expected Behavior
The output order of the tables in the supertibble should reflect the same order in REDCap/the output of `REDCapR::redcap_metadata_read()`. 

# Current Behavior
Currently the order is grouped by `structure` (repeat/nonrepeat) and semi-recognizable ordering, though we don't specify it explicitly anywhere.

# Solution
This should be doable using the `dplyr` ordering functions, identifying the order right after the initial metadata call and then enforcing them at the very end of the `read_redcap_tidy` function.

### Checklist
Before submitting this issue, please check and verify below that the submission meets the below criteria:

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

PR #77, opened 2022-11-11
Fix extract_* Bug

### LINKED ISSUE #75: [BUG] extract_tables no longer works with new field additions
# Expected Behavior
The `extract_table/s` functions should return the expected tables from a `read_redcap_tidy` call.

# Current Behavior
`NULL` values are returned instead of `tibble`s when using `extract` functions with table specs. 

This was not being picked up by our test suite because non-API tests require manual updates of the .RDS files they reference. We may want to think of a better way to update internal/non-API tests or ensure updating the files is a part of our checklist (the current file for doing so is under `/inst/create_test_data.R`).

## How to Reproduce the Bug:
Take a current `read_redcap_tidy` output and supply a specification to `extract_tables`.

## Proposed Solution:
The culprit here is how `extract_tables` shaves off supertibble columns that aren't necessary:
https://github.com/CHOP-CGTDataOps/REDCapTidieR/blob/3944c913990bcd8f9f314a591357c20e01d2adc4/R/extract_table.R#L98-L106

In brief testing I believe all we have to do here is change `select(-"structure)` to `select("redcap_form_name", "redcap_data")`. I think when I originally wrote this it was without the idea that we'd be adding onto the supertibble, but now it makes sense to specifically call out only the columns we need to shore up new additions.

Would welcome thoughts on how to avoid silent test failures like this in the future.

### Checklist

- [x] The issue is atomic

rsh52:
# Description
This PR addresses a minor bug in the `extract_tables` function where originally a `select` statement was made to select everything except one specific row (`structure`). This worked fine before the addition of additional supertibble columns when the original supertibble was only 3 columns. Now it is better to `select` on just the applicable columns necessary for the function to run properly (`redcap_form_name`, `redcap_data`).

> **Note**: This failed silently since our tests rely on .RDS files which were not recreated to reflect the supertibble changes. This should be addressed as noted in #78.

# Proposed Changes 

- Update `extract_tables` with appropriate argument selection
- Recreate .RDS files
- Update `extract` tests since recent change with preservation of table order (#66). 
   - > Should probably write these better to not be dependent on position

### Issue Addressed
**Closes #75**

### PR Checklist

# Code Review

### Code Review Checklist


ezraporter (approved):
This all looks good!

skadauke (approved):

PR #87, opened 2022-11-21
Fix vignette mocks

ezraporter:
This PR fixes the issue with the mocks so that we can run them interactively. It isn't perfect but it will work for now. Some caveats:

1. For some reason the knit button isn't successfully finding out `start-vignette.R` code. I haven't been able to figure out why yet. In the meantime you can run vignettes with `devtools::build_rmd("path_to_vignette")`. That's what you usually want to do anyways since the knit button will use the version of `REDCapTidieR` you have installed rather than the source package you get from `load_all()`.

2. The current setup assumes you have the following in your `.Renviron`:
```
REDCAP_URI = https://redcap.chop.edu/api/
REDCAPTIDIER_CLASSIC_API = [API_KEY]
SUPERHEROES_REDCAP_API = [API_KEY]
REDCAPTIDIER_DEEP_DIVE_VIGNETTE_API = [API_KEY]
```

@skadauke `REDCAPTIDIER_CLASSIC_API` is the same as the one you had called "survey" and  `REDCAPTIDIER_DEEP_DIVE_VIGNETTE_API` is the same as the one you had called "longitudinal"


rsh52 (approved):
Looks and sounds good to me, is this anything we should include in our PR templates?

skadauke (approved):
Looks good! Let's merge so we can see if the vignettes render.

PR #96, opened 2022-11-28
Prefer 'instrument' to 'form' in narrative

skadauke:
# Description

REDCap uses the terms "form" and "instrument" interchangeably. I believe that the word "instrument" has a more specific meaning than "form" and is also the word that data managers would use to describe what a REDCap instrument is. This PR changes "form" to "instrument" in most instances in the docs and outputs. 

### PR Checklist

# Code Review

### Code Review Checklist


R/import_redcap.R
```diff
@@ -88,33 +89,33 @@ import_redcap <- function(redcap_uri,
...
-  # identifiers are kept regardless. This requires adding the form with
-  # identifiers to our redcap_read_oneshot call but filtering out extra,
-  # non-identifier fields in that form
+  # The user may not have requested instrument with identifiers. We need to
+  # make sure identifiers are kept regardless. This requires adding the
+  # instrument with identifiers to our redcap_read_oneshot call but filtering
+  # out extra, non-identifier fields in that instrument
 
-  # Set default forms for API call
+  # Set default instruments for API call
```
rsh52, line 97: Hm, we're changing the comment but leaving the variable below it as `forms_for_api_call`. Should we be changing those as well? It will have a trickle out effect into other variables/functions (like `check_forms_exist` right below).

R/import_redcap.R
```diff
@@ -319,7 +320,7 @@ add_metadata <- function(supertbl, db_metadata, redcap_uri, token) {
     ) %>%
     relocate("field_name", "field_label", "field_type", .before = everything())
 
-  ## Create a record with the project identifier for each form
+  ## Create a record with the project identifier for each instrument
   all_forms <- unique(db_metadata$redcap_form_name)
```
rsh52, line 324: See above comment.

R/labelled.R
```diff
@@ -97,9 +97,13 @@ make_labelled <- function(supertbl, format_labels = NULL) {
...
     redcap_event = "REDCap Event",
     redcap_arm = "REDCap Arm",
+    # TODO: `survey_timestamp` is not a predefined label but it constructed
+    # from [instrument_name]_timestamp. We can either a) change universally
+    # to "survey_timestamp" (which is what we do with `form_status_complete`)
+    # or else we need to change the below.
     survey_timestamp = "REDCap Survey Timestamp",
     redcap_survey_identifier = "REDCap Survey Identifier",
-    form_status_complete = "REDCap Form Status"
+    form_status_complete = "REDCap Instrument Completed?"
```
rsh52, line 106: For the sake of covering all bases, we want to keep this variable as `form_status_complete` and `instrument_status_complete`?

tests/testthat/test-labelled.R
```diff
@@ -151,7 +151,7 @@ test_that("make_labelled applies all predefined labeles", {
     redcap_arm = "REDCap Arm",
     survey_timestamp = "REDCap Survey Timestamp",
     redcap_survey_identifier = "REDCap Survey Identifier",
-    form_status_complete = "REDCap Form Status"
+    form_status_complete = "REDCap Instrument Completed?"
```
rsh52, line 154: See above comment about `form/instrument_status_complete`

rsh52 (changes requested):
Generally looks good, just want to be sure we're clarifying this change everywhere it should be implemented. There are some variables/functions where this could result in crossed wires.

R/import_redcap.R
skadauke: I think the rule should be to use "form" whenever you're programming and interacting with the REDCap API, and use "instrument" whenever you're writing human readable text.

R/labelled.R
skadauke: See above

R/import_redcap.R
rsh52: Sounds good then, in this one case it's a little confusing in just this instance then since the comment and the variable are next to each other.

rsh52 (approved):

Issue #110, opened 2022-12-12
[FEATURE] Add message to unexpected errors with link to bug report


ezraporter:
# Feature Request Description
Unexpected errors should be returned with a message noting that the error wasn't expected by `REDCapTidier` and asking for an issue to be opened on our GitHub.

# Proposed Solution
We can pick up unexpected errors with our `REDCapTidieR_cond` class and wrap exported functions in a `tryCatch()` that looks for non `REDCapTidieR_cond` errors.

## Additional Context
We should do this after #109 since currently errors returned by `checkmate` don't get the `REDCapTidieR_cond` class.

### Checklist

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

ezraporter:
One thought after working on #109. We probably can't reliably assume that all errors not generated by `REDCapTidieR` (i.e. without `REDCapTidieR_cond`) are truly unexpected. For example:

- errors generated by `tidyselect` like `eval_select()` in `extract_tibbles()`
- some errors generated by `REDCapR` like the one for non-hexadecimal tokens
- errors generated by `stringr` which do more sophisticated type checking than we would want to implement

This isn't to say we shouldn't implement this feature but we should think carefully about the message we tag on ("unexpected error" may not cover it) and how many false positives we're likely to generate.

cc: @skadauke @rsh52

ezraporter:
Decided after #107 that we don't want to wrap all exported functions but rather wrap API calls made with `REDCapR`, ex. `try_redcapr()`

PR #116, opened 2022-12-21
Add check_valid_api_token function and tests

### LINKED ISSUE #107: [FEATURE] Improve error message for invalid API token
# Feature Request Description
Currently `REDCapTidieR` fails with an uninformative message when a bad API token is given:

```
Error in `filter()`: 
! Problem while computing `..1 = .data$field_type != "descriptive"`. 
Caused by error in `.data$field_type`: 
! Column `field_type` not found in `.data`. 
Run `rlang::last_error()` to see where the error occurred.
```

Instead we should return an informative message.

# Proposed Solution
We currently pipe the response from `redcap_metadata_read()` into `filter()`:

https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/51ced1e68d20652da0f8854cafeecef88ad2db73/R/read_redcap.R#L76-L81

Instead we can use the information returned by `redcap_metadata_read()` to issue an error message before passing the data on. See `success`, `status_codes`, and `outcome_messages` returned by `redcap_metadata_read()`.

### Checklist

- [x] The issue is atomic

rsh52:
# Description
This PR introduces a simple check function to look for invalid API tokens. Unlike `check_arg_is_valid_token`, which checks for `length` and validity with REDCapR's `sanitize_token`, this function looks at the output of `REDCapR::redcap_metadata_read()` for invalid status return values.

Prior to this, `redcap_metadata_read` failures would give an unhelpful error message related to a `filter` operation we perform immediately after calling the metadata. Now, `check_valid_api_token` will look at the `$success` and `$status_code` values of the metadata list for 403 failure codes.

## Choices and Calls for Input

I decided not to directly capture the error message that `redcap_metadata_read` provides in `$outcome_message` since it's not as friendly as our other error messages, but if we want to pass the exact output message from REDCapR we can:

> "The REDCapR metadata export operation was not successful.  The error message was:\n{\"error\":\"You do not have permissions to use the API\"}"

Otherwise, if we want to make our own informative sub-message as I've made here, happy to incorporate feedback for better language.

# Proposed Changes 

- Introduce `check_valid_api_token` function
- Supply associated tests

### Sample Output

```r
> read_redcap(redcap_uri = redcap_uri
+             , token = paste(sample(1:9, 32, replace = T), collapse = ""))
Error in `read_redcap()`:
✖ The REDCapR metadata export operation was not successful.
ℹ An API token was provided that either does not have the correct privileges or is
  incorrectly formed.
Run `rlang::last_error()` to see where the error occurred.
```

### Issue Addressed
**Closes #107**

### PR Checklist

- [NA] New tests that make API calls use `httptest::with_mock_api` and any new mocks were added to `tests/testthat/fixtures/create_httptest_mocks.R`

# Code Review

### Code Review Checklist


rsh52:
@skadauke implemented as we discussed in the one location. Was thinking about how/if we should add test coverage for this. Trialing "bad" APIs/URIs would require new mocking I imagine, and generally we're looking to capture things we don't know to test for.

Keep me posted, if this is ok I can start applying to the other API calls to start addressing #110

skadauke:
@rsh52 I've been doing some work implementing error messages but am currently getting stuck at the point of trying to re-create mocks. Let's all take a look in the New Year.

skadauke:
Since this PR evolved a bit, here is a point by point summary of changes.

- Rewrite `check_arg_is_valid_token()` to remove `check_arg_is_character()` call and wrap `sanitize_token()` into a `try_fetch()` block to create a pretty `cli_abort()` message
  - Add this `check_arg_is_character()` check explicitly in `read_redcap()`
  - Revise relevant tests in `test-checks.R`
- Wrap `REDCapR::redcap_metadata_read()` into a `try_fetch()` block that does the following:
  - Captures the current call in `call`. This is needed to be able to report the call that caused an error if necessary.
  - Captures the _entire_ output of `redcap_metadata_read()` in `out`. This is needed to throw errors about failure modes that REDCapR doesn't throw errors for
  - Throws an error if `redcap_metadata_read()` completed without erroring out but `out$success == FALSE` indicating the API call was unsuccessful. We want to throw an actual error here.
  - Defines an `error` handler that constructs a message and throws an appropriate error using `cli_abort()`
- Add tests for new error conditions to `test-read_redcap.R`
- Add a mock for a call that leads to a valid REDCap API endpoint rejecting the token

Minor:
- Adjust `.lintr` file
- Style package
- Update `cli_message_examples.R` with new errors


Here is a rendered reprex showing the current error messages for `read_redcap()`:

```r
devtools::load_all(path = "/local/path/to/REDCapTidieR")
#> ℹ Loading REDCapTidieR

options(rlang_backtrace_on_error_report = "none")

# read_redcap

classic_token <- Sys.getenv("REDCAPTIDIER_CLASSIC_API")

## args missing

# read_redcap()

# read_redcap(redcap_uri)

# read_redcap(token = classic_token)

## redcap_uri

read_redcap(123, classic_token)
#> Error in `read_redcap()`:
#> ✖ You've supplied `123` for `redcap_uri` which is not a valid value
#> ! Must be of type 'character', not 'double'

read_redcap(letters[1:3], classic_token)
#> Error in `read_redcap()`:
#> ✖ You've supplied `a`, `b`, `c` for `redcap_uri` which is not a valid
#>   value
#> ! Must have length 1, but has length 3

read_redcap("https://www.google.com", classic_token)
#> Error in `read_redcap()`:
#> ✖ The REDCapR metadata export operation was not successful.
#> ! The URL returned the HTTP error code 405 (POST Method not allowed).
#> ℹ The most likely reason is that the provided URI does not point to a REDCap
#>   API endpoint.
#> ℹ URI: `https://www.google.com`

read_redcap("https://www.google.comm", classic_token)
#> Error in `read_redcap()`:
#> ✖ The REDCapR metadata export operation was not successful.
#> ! Could not resolve the hostname.
#> ℹ The most likely reason is that the provided URI is incorrect.
#> ℹ URI: `https://www.google.comm`

## token

read_redcap(redcap_uri, 123)
#> Error in `read_redcap()`:
#> ✖ You've supplied `123` for `token` which is not a valid value
#> ! Must be of type 'character', not 'double'

read_redcap(redcap_uri, letters[1:3])
#> Error in `read_redcap()`:
#> ✖ You've supplied `a`, `b`, `c` for `token` which is not a valid value
#> ! Must have length 1, but has length 3

read_redcap(redcap_uri, "")
#> Error in `read_redcap()`:
#> ✖ The token is an empty string, not a valid 32-character hexademical
#>   value.
#> ℹ API token: ``

read_redcap(redcap_uri, "CC0CE44238EF65C5DA26A55DD749AF7") # 31 hex characters
#> Error in `read_redcap()`:
#> ✖ The token is not a valid 32-character hexademical value.
#> ℹ API token: `CC0CE44238EF65C5DA26A55DD749AF7`

read_redcap(redcap_uri, "CC0CE44238EF65C5DA26A55DD749AF7A") # will be rejected by server
#> Error in `read_redcap()`:
#> ✖ The REDCapR metadata export operation was not successful.
#> ! You supplied an API token that either does not have the correct privileges or
#>   is incorrectly formed.
#> ℹ API token: `CC0CE44238EF65C5DA26A55DD749AF7A`

## raw_or_label

read_redcap(redcap_uri, classic_token, raw_or_label = "bad option")
#> Error in `read_redcap()`:
#> ✖ You've supplied `bad option` for `raw_or_label` which is not a valid
#>   value
#> ! Must be element of set {'label','raw'}, but is 'bad option'

## forms

read_redcap(redcap_uri, classic_token, forms = 123)
#> Error in `read_redcap()`:
#> ✖ You've supplied `123` for `forms` which is not a valid value
#> ! Must be of type 'character' (or 'NULL'), not 'double'

## export_survey_fields

read_redcap(redcap_uri, classic_token, export_survey_fields = 123)
#> Error in `read_redcap()`:
#> ✖ You've supplied `123` for `export_survey_fields` which is not a valid
#>   value
#> ! Must be of type 'logical', not 'double'

read_redcap(redcap_uri, classic_token, export_survey_fields = c(TRUE, TRUE))
#> Error in `read_redcap()`:
#> ✖ You've supplied `TRUE`, `TRUE` for `export_survey_fields` which is not
#>   a valid value
#> ! Must have length 1, but has length 2

## suppress_redcapr_messages

read_redcap(redcap_uri, classic_token, suppress_redcapr_messages = 123)
#> Error in `read_redcap()`:
#> ✖ You've supplied `123` for `suppress_redcapr_messages` which is not a
#>   valid value
#> ! Must be of type 'logical', not 'double'

read_redcap(redcap_uri, classic_token, suppress_redcapr_messages = c(TRUE, TRUE))
#> Error in `read_redcap()`:
#> ✖ You've supplied `TRUE`, `TRUE` for `suppress_redcapr_messages` which
#>   is not a valid value
#> ! Must have length 1, but has length 2

# bind_tibbles

bind_tibbles(123)
#> Error in `bind_tibbles()`:
#> ✖ You've supplied `123` for `supertbl` which is not a valid value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

supertbl <- tibble(redcap_data = list())
bind_tibbles(supertbl, environment = "abc")
#> Error in `bind_tibbles()`:
#> ✖ You've supplied `<tibble[,1]>` for `supertbl` which is not a valid
#>   value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

bind_tibbles(supertbl, tbls = 123)
#> Error in `bind_tibbles()`:
#> ✖ You've supplied `<tibble[,1]>` for `supertbl` which is not a valid
#>   value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

# extract_tibbles

extract_tibbles(letters[1:10])
#> Error in `extract_tibbles()`:
#> ✖ You've supplied `a`, `b`, `c`, …, `i`, `j` for `supertbl` which is not
#>   a valid value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

# extract_tibble

extract_tibble(123, "my_tibble")
#> Error in `extract_tibble()`:
#> ✖ You've supplied `123` for `supertbl` which is not a valid value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

supertbl <- tibble(redcap_data = list()) %>%
  as_supertbl()
extract_tibble(supertbl, tbl = 123)
#> Error in `extract_tibble()`:
#> ✖ You've supplied `123` for `tbl` which is not a valid value
#> ! Must be of type 'character', not 'double'

extract_tibble(supertbl, tbl = letters[1:3])
#> Error in `extract_tibble()`:
#> ✖ You've supplied `a`, `b`, `c` for `tbl` which is not a valid value
#> ! Must have length 1, but has length 3

# make_labelled

make_labelled(123)
#> Error in `make_labelled()`:
#> ✖ You've supplied `123` for `supertbl` which is not a valid value
#> ! Must be of class <redcap_supertbl>
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

missing_col_supertbl <- tibble(redcap_data = list()) %>%
  as_supertbl()
make_labelled(missing_col_supertbl)
#> Error in `make_labelled()`:
#> ✖ You've supplied `<rdcp_spr[,1]>` for `supertbl` which is not a valid
#>   value
#> ! Must contain `supertbl$redcap_metadata`
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()`

missing_list_col_supertbl <- tibble(redcap_data = list(), redcap_metadata = 123) %>%
  as_supertbl()
make_labelled(missing_list_col_supertbl)
#> Error in `make_labelled()`:
#> ✖ You've supplied `<rdcp_spr[,2]>` for `supertbl` which is not a valid
#>   value
#> ! `supertbl$redcap_metadata` must be of type 'list'
#> ℹ `supertbl` must be a REDCapTidieR supertibble, generated using
#>   `read_redcap()` 
```

R/read_redcap.R
```diff
@@ -72,20 +73,83 @@ read_redcap <- function(redcap_uri,
...
-    redcap_uri = redcap_uri,
-    token = token,
-    verbose = !suppress_redcapr_messages
-  )$data %>%
+  # Capture unexpected metadata API call errors
+  try_fetch(
+    {
+      call <- current_call()
+
+      out <- redcap_metadata_read(
```
ezraporter, line 90: I think this would be more consistent with the flow of `read_redcap` since we introduce an `out` object on L220
```suggestion
      db_metadata <- redcap_metadata_read(
```

R/read_redcap.R
```diff
@@ -72,20 +73,83 @@ read_redcap <- function(redcap_uri,
...
+    },
+    error = function(cnd) {
+      error_message <- c("x" = "The REDCapR metadata export operation was not successful.")
+
+      # Show original error?
+      parent <- NULL
+
+      error_class <- c("metadata_api_call_failed", "REDCapTidieR_cond")
+
+      if (exists("out") && out$status_code == 403) {
```
ezraporter, line 112: This seems slightly fragile since we're looking outside of the current function env for an object called `out`. An alternative would be to:

1. Pass `out$status_code` as error metadata on L98-102 like

```r
abort(
    class = c("redcapr_api_call_success_false", "REDCapTidieR_cond"),
    status_code = out$status_code
)
```

Then you could safely reference it as `cnd$status_code`

2. Move L112-125 to a `try_fetch` handler for `redcapr_api_call_success_false` above the one for `error` so the code would look like:

```r
try_fetch({
    ...
    },
    redcapr_api_call_success_false = function(cnd) {
        # L112-125 logic
        if (cnd$status_code == 403) {
        ...
        } else if ...
    },
    error = function(cnd) {
        # L125-139 logic
    }
)

**Review by ezraporter — CHANGES_REQUESTED** [review]
This looks really good overall. A couple suggestions to consider in the `try_fetch` block to refine how we're passing around the status code. I tried to explain what I was thinking but if it sounds like a good idea and it isn't totally clear how to implement I can try to mock it up.

I also think it would keep `read_redcap()` a little neater if we move the handler code to a named function at the bottom of `read_redcap.R` rather than keeping it as an anonymous function. Best solution is probably to put all the `try_fetch` logic in a helper function but that's a little more complicated.

File: R/read_redcap.R (hunk 2 above)
Comment (skadauke) [reply R/read_redcap.R:112]: How awesome, I had no idea that you can pass arbitrary metadata inside the `cnd` object! Implemented mostly as suggested, except I kept all the error handling in one function instead of breaking into two, for now, for simplicity's sake, but it's a great point that we can handle each error class with a separate handler function. Anyhow, I think we can later decide how to refactor.

File: R/checks.R [hunk 3]
```diff
@@ -382,10 +381,19 @@ check_arg_choices <- wrap_checkmate(check_choice)
 check_arg_is_valid_token <- function(x,
```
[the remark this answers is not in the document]
skadauke: Done!

R/read_redcap.R
skadauke: Done!

rsh52:
Generally like all the changes, and the return error messages look succinct and informative.

> I also think it would keep `read_redcap()` a little neater if we move the handler code to a named function at the bottom of `read_redcap.R` rather than keeping it as an anonymous function. Best solution is probably to put all the `try_fetch` logic in a helper function but that's a little more complicated.

I agree this should be considered, especially as we're about to apply this across every REDCapR API call and should minimize duplication of large code blocks to address #110. We can do the refactor after this PR.

PR #122, opened 2023-02-02
Lint in test suite

ezraporter:
This PR adds the lint check to our test suite with `expect_lint_free()` and remove the GHA

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,6 +34,7 @@ Suggests:
     httptest,
     knitr,
     labelled,
+    lintr,
     readr,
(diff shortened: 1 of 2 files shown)


rsh52:
Looks great, love a PR that only adds 4 lines of code. Only thing was didn't Stephan say he wanted to keep both the test as well as the Actions lintr?

ezraporter:
> Looks great, love a PR that only adds 4 lines of code. Only thing was didn't Stephan say he wanted to keep both the test as well as the Actions lintr?

Oh I didn't catch that. I'll add it back

PR #139, opened 2023-03-14
Edit vignettes

skadauke:
# Description

Edits to the README and vignettes

### DIFF
--- a/vignettes/diving_deeper.Rmd
+++ b/vignettes/diving_deeper.Rmd
@@ -41,7 +41,7 @@ In the [Getting Started](REDCapTidieR.html) vignette, we stated that importing r
 
 The `REDCapR::redcap_read_oneshot()` function pulls data from a REDCap project via the [REDCap API](glossary.html#redcap-api) and provides it to the R environment without performing much processing. It returns a list with an element named `data` that contains all of the data of the project as a single data frame.
 
-Below, we use this function to import data from the [Superhero REDCap database](REDCapTidieR.html#case-study-the-superhero-database), which contains a [**nonrepeating**](glossary.html#nonrepeating-instrument) instrument `heroes_information` and a [**repeating**](glossary.html#repeating-instrument) instrument `super_hero_powers`. We use `rmarkdown::paged_table()` to allow you to explore this large data frame in the browser.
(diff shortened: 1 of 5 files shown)


skadauke:
Hi, I'm reviewing the vignettes, and I noticed the following:

image

Is this still true? I thought we got rid of this? Are we throwing a warning when we detect this?

Also, I don't think the following is accurate? Rather, REDCap doesn't actually allow you to configure a repeating instrument in a repeating event, correct?

image

rsh52:
> Hi, I'm reviewing the vignettes, and I noticed the following:
> 
> image
> 
> Is this still true? I thought we got rid of this? Are we throwing a warning when we detect this?

This is true, the [`check_repeat_and_nonrepeat`](https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/a2706d116c0d37c8a3b46bb27cb3e35bc35e0c7d/R/checks.R#L113) function is dedicated to detecting it. We still want this to exist since it should never happen.
 
> Also, I don't think the following is accurate? Rather, REDCap doesn't actually allow you to configure a repeating instrument in a repeating event, correct?
> 
> image

Correct, this isn't possible to do from the REDCap UI. I thought we had a check for this regardless, but I might be mistaken.

vignettes/diving_deeper.Rmd
```diff
@@ -178,13 +182,13 @@ hematology |>
   rmarkdown::paged_table()
 ```
 
-Note that both `physical_exam` and `hematology` are **nonrepeating** instruments. REDCapTidieR **only** allows nonrepeating instruments to occur in repeating events. It will throw an error if it detects a repeating instrument associated with a repeating event.
+Note that REDCapTidieR allows for an instrument to be associated with both repeating and nonrepeating events at the same time. In this case, the the `redcap_event_instance` column will be `NA` in rows that correspond with data captured as part of a nonrepeating event.
```
rsh52, line 185: I think we still need to clarify that this is only allowed for nonrepeating instruments, this sounds like it's saying any instrument can be associated with both event structures.

rsh52 (commented):
Generally looks good, one comment. 👍

vignettes/diving_deeper.Rmd
skadauke: Hm, I think we're making this point at least 3 times in other places. Adding it here I think will make things less clear.

rsh52: I see no harm in a 4th for consistency, up to you though 👍

PR #146, opened 2023-04-05
Small mocks fix

ezraporter:
This PR fixes an issue where creating mocks fails when you don't have **all** of the credentials in `fake_credentials.csv` in your `.Renviron`. It's really only an issue when rerunning the vignettes where you may want to regenerate mocks for a vignette that uses only a subset of the test REDCaps without ensuring your `.Renviron` is totally up to date.

### DIFF
--- a/inst/httptest/redact.R
+++ b/inst/httptest/redact.R
@@ -8,7 +8,16 @@ function(response) {
   # Replace each token with a fake token
   purrr::reduce(
     names(fake_creds),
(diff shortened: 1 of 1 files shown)


ezraporter:
FYI @rsh52 this is an issue I noticed and wrote a fix for when I was looking at the mocks failure earlier this week. Forgot I never pushed it to your guess max PR

rsh52 (approved):
LGTM thanks for following up on and doing this!

PR #153, opened 2023-05-24
Add Data Access Group Support

### LINKED ISSUE #57: [FEATURE] REDCapTidieR Should Support Data Access Groups
# Feature Request Description
`REDCapTidieR` should support the inclusion of Data Access Groups (DAGs). This is available in `REDCapR`'s [`export_data_access_groups`](https://ouhscbbmc.github.io/REDCapR/reference/redcap_read_oneshot.html#arguments-1) argument.

## Additional Context
Data Access Groups are used for multiple collaborators and make it possible to changes access rights. The `redcap_data_access_group` variable comes out in the database export as its own column, but does not come out in the metadata (similar to `redcap_repeat_instrument`/`instance`).

> This could be a good use of a R/Medicine hackathon topic

### Potential Problems

It will be necessary to determine what happens to a user with limited DAG access rights.

**Update 2023-03-28:**
- This should be solved using a `redcap_dags` argument (by default set to `FALSE`). Likely just needs to be incorporated with the `REDCapR::redcap_read_oneshot` argument for [`export_data_access_groups`](https://ouhscbbmc.github.io/REDCapR/reference/redcap_read_oneshot.html)

### Checklist

- [x] The issue is atomic

rsh52:
# Description
This PR provides support for Data Access Groups (DAGs) primarily by introducing a new argument to `read_redcap()` in the form of `export_data_access_groups` which plugs directly into our `REDCapR::redcap_read_oneshot()` call (and is identical to the argument used there).

Unlike in `redcap_read_oneshot()`, I opted to make the default `TRUE` to both follow suit with `export_survey_fields` and because there's nothing to really lose if no DAGs are found. Instead, DAGs will be auto-added if they are detected and the same behavior exists in `REDCapR` (although the default there is `FALSE`).

To supplement this, two new REDCapTidieR test databases were made using DAGs for a non-longitudinal and a longitudinal database. Mocks were also updated, so you may want to add these environ vars to your files once you generate your own API keys (see `/inst/misc/fake_credentials` for names).

# Proposed Changes 

- Add support for DAGs to `read_redcap()` and updated associated `clean_*`/`distill_*` functions
   - This is very similar to logic we already included for survey fields
- Add `redcap_data_access_group`/"REDCap Data Access Group" to labels
- Update mocks and vignette mocks
- Add tests using new mocks to confirm new column and label
   - I think these could be made better and am open to suggestions!

### Issue Addressed
**Closes #57**

### PR Checklist

# Code Review

### Code Review Checklist


R/clean_redcap.R
```diff
@@ -163,6 +168,7 @@ distill_nonrepeat_table <- function(form_name,
       any_of(c("redcap_event", "redcap_arm", "redcap_form_instance", "redcap_event_instance")),
       .after = all_of(my_record_id)
     ) %>%
+    relocate(any_of("redcap_data_access_group"), .after = all_of(my_record_id)) %>%
```
skadauke, line 171: So `redcap_data_access_group` will come after `record_id`? Hm, I'm wondering if that's the best place. My gut feeling is that pushing this all the way to the right would be better, into the proximity of the survey identifiers?

R/clean_redcap.R
```diff
@@ -233,13 +239,19 @@ distill_repeat_table <- function(form_name,
...
+  }
+
   db_data %>%
     add_partial_keys() %>%
     filter(
       !is.na(.data$redcap_form_instance) &
         .data$redcap_repeat_instrument == my_form
     ) %>%
     select(all_of(my_fields), "redcap_form_instance") %>%
+    relocate(any_of("redcap_data_access_group"), .after = all_of(my_record_id)) %>%
```
skadauke, line 254: See comment above

R/clean_redcap_long.R
```diff
@@ -191,6 +196,7 @@ distill_nonrepeat_table_long <- function(form_name,
       any_of(c("redcap_event", "redcap_arm", "redcap_form_instance", "redcap_event_instance")),
       .after = !!my_record_id
     ) %>%
+    relocate(any_of("redcap_data_access_group"), .after = all_of(my_record_id)) %>%
```
skadauke, line 199: See above

R/clean_redcap_long.R
```diff
@@ -311,6 +322,7 @@ distill_repeat_table_long <- function(form_name,
...
```
skadauke, line 325: See above

R/labelled.R
```diff
@@ -91,6 +91,7 @@ make_labelled <- function(supertbl, format_labels = NULL) {
   data_labs <- c( # nolint: object_usage_linter
```
skadauke, line 91: `labels` :)

R/read_redcap.R
```diff
@@ -75,6 +78,7 @@ read_redcap <- function(redcap_uri,
                         raw_or_label = "label",
                         forms = NULL,
                         export_survey_fields = TRUE,
+                        export_data_access_groups = TRUE,
```
skadauke, line 81: Wondering, should `export_survey_fields` and `export_data_access_groups` be `TRUE` or `NULL`, where `NULL` means that export them if they are there?

R/read_redcap.R
```diff
@@ -75,6 +78,7 @@ read_redcap <- function(redcap_uri,
...
                         suppress_redcapr_messages = TRUE,
                         guess_max = .Machine$integer.max) {
   check_arg_is_character(redcap_uri, len = 1, any.missing = FALSE)
```
skadauke, line 84: Need a `check_arg_` for `export_data_access_groups`

tests/testthat/test-read_redcap.R
```diff
@@ -565,3 +565,48 @@ test_that("read_redcap works for a large sparse database", {
     sapply(class) %>%
     expect_equal(expected_col_types)
 })
+
+test_that("read_redcap works with non-longitudinal Data Access Groups", {
+  httptest::with_mock_api({
```
skadauke, line 570: Do we want to check that a non-DAG database doesn't return the `redcap_data_access_group` column?

tests/testthat/test-read_redcap.R
skadauke, line 1: I think we need to add an arg check

skadauke (changes requested):
Can you also add an entry for Data Access Groups to the Glossary?

R/clean_redcap.R
rsh52: In REDCapR it comes after the `record_id`, `redcap_repeat_instance`, `redcap_repeat_instrument`, `redcap_repeat_instance` columns:

```r
> redcapr_out_dag <- REDCapR::redcap_read_oneshot(redcap_uri, token, export_data_access_groups = TRUE)$data %>% tibble()
6 records and 8 columns were read from REDCap in 1.1 seconds.  The http status code was 200.                                           
> redcapr_out_dag
# A tibble: 6 × 8
  record_id redcap_repeat_instrument redcap_repeat_instance redcap_data_access_group
      <dbl> <chr>                                     <dbl> <chr>                   
1         1 NA                                           NA dag1                    
2         1 repeat_form_1                                 1 dag1                    
3         1 repeat_form_1                                 2 dag1                    
4         2 NA                                           NA dag2                    
5         3 NA                                           NA dag3                    
6         4 NA                                           NA NA                      
# ℹ 4 more variables: non_repeat_form_1_txt <lgl>, non_repeat_form_1_complete <dbl>,
#   repeat_form_1_txt <dbl>, repeat_form_1_complete <dbl>
```

Personally after working with REDCapR a lot prior to our package and in the context of DAGs, I'd prefer to see it in a similar place and not at the end of the dataset. I think this makes sense since from an analyst point of view this an additional identifier to work off of, even if its not a part of the composite key.

R/labelled.R
rsh52: Gah!
👍

rsh52: I'll just go ahead and change all of the `*_labs` then, this was in place since `make_labelled()` was first written.

R/read_redcap.R
rsh52: This follows the REDCapR conventions again, I'm impartial but there is some sense in keeping in line with REDCapR since these get directly copied from it (although in REDCapR they're default `FALSE`).

tests/testthat/test-read_redcap.R
rsh52: Sure!

R/read_redcap.R
skadauke: The only reason would be to fail more loudly if someone expected DAGs (or survey fields), intentionally set the field to `TRUE` and then receives an informative error message if the expected DAG or survey fields aren't there. Admittedly this is probably a rare edge case but on the other hand I think it would be good to implement a consistent implementation of these sorts of flags across all functions of the package. Might be worth discussing at standup.

R/clean_redcap.R
skadauke: Thanks for the additional info. I would argue that we are already taking some liberties with changing the non-data column structure to make it more logical. As of now, REDCapTidieR logic is that you have the composite key, then data columns, and then row-level metadata columns. I think DAG is a row-level metadata column akin to the survey identifier or form completion question. Might be worth discussing at standup.

R/read_redcap.R
rsh52: TODO: implement default is `NULL` which returns if detected, fails if specified `TRUE` and not detected.

R/clean_redcap.R
rsh52: Noting from today's meeting, after discussion opting to keep DAG after composite key identifiers. Sent a message to Paul as well for general feedback.

R/checks.R
```diff
@@ -557,3 +558,48 @@ check_arg_is_valid_extension <- function(x,
...
+#' @param db_data The REDCap database output generated by
+#' \code{REDCapR::redcap_read_oneshot()$data}
+#' @param col The column to check for in `redcap_data`
+#' @param arg The argument used for the column check
+#' @param call The calling environment to use in the error message
+#'
+#' @keywords internal
+check_data_arg_exists <- function(db_data, col, arg, call = caller_env()) {
+
+  msg_x <- "Invalid request: {.arg {arg}}."
```
skadauke, line 592: I think this error message could be a bit more helpful. If you're trying to pull survey fields from a project that doesn't have any survey instruments, it should say so. And if you're trying to pull the DAG column from a non-DAG enabled project it should say so. Also, I wouldn't say "identifying column" because we use that term to describe the partial key columns.

R/read_redcap.R
```diff
@@ -170,6 +188,23 @@ read_redcap <- function(redcap_uri,
   # Check that results were returned
   check_redcap_populated(db_data)
 
+  # Check that requested args exist
```
skadauke, line 191: How about: 
`# Check whether DAGs or survey fields are present in the project, if requested`

tests/testthat/test-read_redcap.R
```diff
@@ -565,3 +579,69 @@ test_that("read_redcap works for a large sparse database", {
...
+  # Check for expected label
+  out_dag_long_labelled <- out_dag_long %>% make_labelled()
+
+  dag_label_long <- labelled::lookfor(out_dag_long_labelled$redcap_data[[1]])
+  dag_label_long <- dag_label_long$label[dag_label_long$variable == "redcap_data_access_group"]
+
+  expect_equal(dag_label_long, c("redcap_data_access_group" = "REDCap Data Access Group"))
+})
+
+test_that("read_redcap fails for args that don't exist but are requested", {
```
skadauke, line 628: How about: "read_redcap fails if DAG or survey columns are explicitly requested but don't exist"

tests/testthat/test-read_redcap.R
skadauke: I think this still needs to be done?

skadauke (changes requested):

tests/testthat/test-read_redcap.R
rsh52: Odd, somehow it seems to have gotten lost from [this commit (line628)](https://github.com/CHOP-CGTInformatics/REDCapTidieR/pull/153/commits/e2dfb64825fa5546b398676ae4f4e817868ca166). I'll add it back in.

R/deprecated.R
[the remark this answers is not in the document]
rsh52: I had to change them to get the checks to pass, otherwise the deprecation tests would constantly fail. Is it typical to get rid of them during 0.* releases, or is it more kosher to wait for 1.0?

R/checks.R
rsh52: Sure, makes sense to me

tests/testthat/test-read_redcap.R
rsh52: Ah nvm, this was me. Added it back in but needed a slight rework for the check.

R/deprecated.R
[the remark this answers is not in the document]
rsh52: Note: Decided to remove deprecated functions entirely. Done in recent commit.

skadauke (approved):
Looks great to me!

PR #156, opened 2023-06-01
Improve `REDCapR` error messages

ezraporter:
# Description
This PR improves error messages produced by `try_redcapr()` in 2 scenarios:

1. The `REDCapR` operation returns with `success == FALSE` but doesn't error

In these cases we check for a few common error codes

https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/fe147b4e476c496f5ded1dfc54b4a66601ce53fc/R/utils.R#L641-L657

and otherwise issue a generic "unexpected error" message with our bug report link. This hides information `REDCapR` may have returned in `outcome_message`.

This PR includes the contents of `outcome_message` in the error. See screenshot 1.

2. `REDCapR` returns `success == FALSE` due to a deleted project

This is the situation that instigated investigating scenario 1. I added an explicit check for this error message to the error code checks above. 

# Proposed Changes 

- Add the content of `outcome_message` as the `parent` in `abort` when handling an unexpected error
- Add explicit check for deleted projects

**Screenshots**
New unexpected error message:
(Note that this won't actually ever appear because we added the explicit check for this case)
Screenshot 2023-06-01 at 3 19 59 PM

New deleted project error message:
Screenshot 2023-06-01 at 3 20 29 PM

### PR Checklist
Before submitting this PR, please check and verify below that the submission meets the below criteria:

- [x] New/revised functions have associated tests
- [x] New/revised functions that update downstream outputs have associated static testing files (`.RDS`) updated under `inst/testdata/create_test_data.R`
- [x] New tests that make API calls use `httptest::with_mock_api` and any new mocks were added to `tests/testthat/fixtures/create_httptest_mocks.R`
- [x] New/revised functions use appropriate naming conventions
- [x] New/revised functions don't repeat code
- [x] Code changes are less than **250** lines total
- [x] Issues linked to the PR using [GitHub's list of keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- [x] The appropriate reviewer is assigned to the PR
- [x] The appropriate developers are assigned to the PR
- [x] Pre-release package version incremented using `usethis::use_dev_version()`

# Code Review
This section to be used by the reviewer and developers during Code Review after PR submission

### Code Review Checklist

- [ ] I checked that new files follow naming conventions and are in the right place
- [ ] I checked that documentation is complete, clear, and without typos
- [ ] I added/edited comments to explain "why" not "how"
- [ ] I checked that all new variable and function names follow naming conventions
[...]

PR #159, opened 2023-07-17
Fix `REDCapR` revdep failure

ezraporter:
# Description
This PR regenerates the `httptest` mocks to be compatible with the latest version of `REDCapR`. Currently our `renv.lock` points towards the dev version of `REDCapR` but I imagine we won't push this until `REDCapR 1.2.0` is on CRAN at which point we'll change that.

This also removes the custom error message for deleted REDCap projects introduced in #156. When regenerating the mocks I was not able to reproduce this error message with `REDCapR 1.1.0` or `1.2.0`.  My guess is that this is due to updates to the CHOP REDCap instance which means checking for deleted as opposed to nonexistent REDCap projects is likely too fragile.

# Proposed Changes 

- Regenerate mocks
- Remove custom error message for deleted REDCap projects and associated tests

### PR Checklist

# Code Review

### Code Review Checklist


rsh52:
> This also removes the custom error message for deleted REDCap projects introduced in https://github.com/CHOP-CGTInformatics/REDCapTidieR/pull/156. When regenerating the mocks I was not able to reproduce this error message with REDCapR 1.1.0 or 1.2.0. My guess is that this is due to updates to the CHOP REDCap instance which means checking for deleted as opposed to nonexistent REDCap projects is likely too fragile.

So what is the behavior now when a deleted project is called? The tryCatch generic message?

ezraporter:
> > This also removes the custom error message for deleted REDCap projects introduced in #156. When regenerating the mocks I was not able to reproduce this error message with REDCapR 1.1.0 or 1.2.0. My guess is that this is due to updates to the CHOP REDCap instance which means checking for deleted as opposed to nonexistent REDCap projects is likely too fragile.
> 
> So what is the behavior now when a deleted project is called? The tryCatch generic message?

It will probably depend on your institution's REDCap instance but for CHOP it's returning `403` so we get this one: https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/82c8b8a67cdf59db9ac492208669e5218ad87553/R/utils.R#L645-L651

If I recall correctly, we were previously getting `404` with the message `The REDCap project no longer exists because it was deleted` although I can't recreate that to be sure. If a user gets that response they'll get that `The REDCap project no longer exists...` message in the error from `REDCapTidieR`.

rsh52:
Got it thanks, I forgot we had updated the error message defaults to return what the REDCap API returns. In that case I think this should be good, in general it _should_ be relatively rare that a `read_redcap()` call gets made against a deleted database. Only time I can see it popping up is for routine calls that are left on a schedule.

See what Will says but otherwise I think this looks fine.

rsh52 (approved):
I think this looks good, I would say merge it once checks finish passing and then make sure that the pkgdown site and vignettes look right on the pkgdown site that GitHub makes. You may also want to include an extra note in the CRAN comments for more explanation of this update since the NEWS is a little vague (but that's fine for users). We can submit tomorrow.

Issue #160, opened 2023-09-08
[BUG] openxlsx2 Deprecation Warning


rsh52:
# Current Behavior
Less a bug and more of an alert. After a recent `openxlsx2` update, a warning now appears as below for `write_redcap_xlsx()`:

```
Warning: 'wb_save(path)' is deprecated.
Use 'wb_save(file)' instead.
See help("Deprecated") and help("openxlsx2-deprecated").
```

We need to update the function to remove this warning.

### Checklist
Before submitting this issue, please check and verify below that the submission meets the below criteria:

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [ ] Developers are assigned to the issue
- [x] Labels are assigned to the issue

PR #162, opened 2023-09-12
Update to v0.4.1.9000

rsh52:
Minor update for the dev package version and removal of some comments per #160.

In testing, updating `convert_perc` to `openxlsx2::wb_add_ignore_error()` doesn't gain us much and makes the purpose a little more confusing. 

Removing handling of sheet value names using the combination of `str_trunc()` (to handle > 31 characters), punctuation removal, and whitespace removal caused failures for some complex names from our test databases. I think it's better to mandate some rules up front that ensures stability later, since it seems removal of our existing rules results in > 31 character errors and mishandling of "/" characters.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: REDCapTidieR
 Type: Package
 Title: Extract 'REDCap' Databases into Tidy 'Tibble's
-Version: 0.4.1
+Version: 0.4.1.9000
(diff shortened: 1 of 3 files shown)


olivroy:
oops, sorry for bothering, I just typed that quickly. 

Feel free to file issues or festure requests! openxlsx2 is still under active dev!

PR #170, opened 2024-01-26
JOSS Feedback Round 1

### LINKED ISSUE #167: Please update redcapAPI reference
In the JOSS paper, please update the redcapAPI reference. In review I realized the existing reference one was out of date, and due to a few changes of ownership we have no control over the old DOI. I apologize, but this has the benefit that it's now as update to date as I can make it. 

The new one is: https://zenodo.org/records/10564837

### LINKED ISSUE #168: Block Matrix line 48 paper
The existing paper on line 48 states '...only `REDCapTidieR` restructures the block matrix into an easy to use format ...'

This was true a true statement a year ago, but the package `redcapAPI` at present supports splitting out the block sparse matrix into separate data.frames with a single call `exportBulkRecords`. 

The paper's description of the block sparse problem is excellent. 

A big dividing line between the `REDCapTidierR` and `redcapAPI` is that `redcapAPI` is by design intended to be mostly base R, whereas `REDCapTidieR` has full `tibble` support which a lot of users love.

rsh52:
# Description
This PR addresses requests from the first round of reviewer feedback:

- Update the `redcapAPI` citation
- Update claims about `REDCapTidieR` and `redcapAPI` capabilities and specify differences
- Update acknowledgements to align with authorship

The original paper word count according to the JOSS bot was 1189 words (limit in the criteria is supposed to be 1000), and this adds even more. If there's any feedback on things to remove, happy to implement.

This PR closes #168 and closes #167


rsh52:
For the time being placed the failing snapshot test in a `skip_on_ci()` so this doesn't block JOSS review. We can fix the test with a similar approach to the `CGTTrialsReporter` at a later point.

@spgarbet I believe you already "approved" on the review page, but tagging you for awareness as well.

ezraporter (approved):
Changes look good to me!

skadauke (approved):
LGTM!

PR #126, opened 2023-02-10
Repeating Events Support

rsh52:
# Description
This PR looks to enable appropriate handling of repeat events in REDCap as outlined in #120. The default behavior related to REDCap repeat event exports and the method to handle them via the composite key table are both outlined below under "**Screenshots**".

There are a few known problems with this PR lending to failed actions:
- ~~Warnings appear when building the test suite, but have not been reproducible in local code or individual testing. It seems to be related to `tidyselect` semantics but is unclear why this is happening~~
- ~~The recently added mock isn't firing correctly on GitHub Actions, maybe you can take another look @ezraporter?~~

I also made a few comments on follow-up to do's related to updating `make_labelled` and the vignettes. For now I would just like to get feedback on the new methods for handling events and perhaps you guys will find something I wrote that resolves these issues.

# Proposed Changes 
List changes below in bullet format:

- Change appropriate instances of `redcap_repeat_instance` and convert to `redcap_form_instance` and `redcap_event_instance`
   - **Note**: This is a _**breaking change**_ for older versions of the package
   - "Appropriate" instances meaning that `add_partial_keys` takes place _after_ other operations in the package use `redcap_repeat_instance`, so not all are 1:1 replaced.
- Create new `create_repeat_instance_vars` to handle above conversion and reside in `add_partial_keys` function
- Update `clean_redcap` and `clean_redcap_long` functions to process new outputs of `add_partial_keys`
- Update associated mocks, tests, and documentation

#### Sample REDCap Repeat Events Default Output:

record_id | redcap_event_name | redcap_repeat_instrument | redcap_repeat_instance | nr_instr_field | nr_instrument_complete | r_instr_field | r_instrument_complete
-- | -- | -- | -- | -- | -- | -- | --
1 | nr_event_arm_1 | NA | NA | test | 0 | NA | NA
1 | nr_event_arm_1 | r_instrument | 1 | NA | NA | test | 0
1 | nr_event_arm_1 | r_instrument | 2 | NA | NA | test2 | 0
1 | nr_event_arm_1 | r_instrument | 3 | NA | NA | NA | 0
1 | r_event_arm_1 | NA | 1 | test | 0 | NA | NA
1 | r_event_arm_1 | NA | 2 | test | 0 | NA | NA
2 | nr_event_arm_1 | NA | NA | test | 0 | NA | NA
3 | nr_event_arm_1 | NA | NA | NA | 0 | NA | NA
[...]

PR #177, opened 2024-03-01
Add Mixed Repeat/Nonrepeat Instrument Support

rsh52:
# Description
This PR seeks to add a new parameter to `read_redcap()` which will allow users to override the check in place that stops data exports for REDCap projects where instruments are detected to be both repeating and nonrepeating. The background rationale for not allowing this was to keep to tidy data principles.

For the purposes of this PR and documentation, here are my definitions for some things:

- **Mixed Structure**: One of the supertibble columns is called `structure`, and has only ever been "nonrepeating" / "repeating" for values. I wanted to keep this terminology so some of the updated functions use "structure" / "mixed structure"
- **Mixed Instrument**: An instrument that has a mixed structure

# Proposed Changes 

- Add param `enable_repeat_nonrepeat` to `read_readcap()` and lower-level function `clean_redcap_long()`
- Move shared logic in `check_repeat_and_nonrepeat()` to new function `get_mixed_structure_fields()`
- Wrap `get_mixed_structure_fields()` in handler function for converting nonrepeat parts of mixed structure instruments to repeating ones with a single instance via `convert_mixed_instrument()`
- Update tests
- Update `test_creds.R` (We now have a new database where a new env variable is needed: `REDCAPTIDIER_MIXED_STRUCTURE_API`)
- Update NEWS and version (currently considering this a minor update to 1.1.0)

# Remaining TODOs:

- [ ] Update Diving Deeper vignette/article (holding off until we're happy with logic choices and syntax)

### Issue Addressed
**Addresses #126**
**Addresses #169**

### PR Checklist

- [ x New/revised functions have associated tests

# Code Review

### Code Review Checklist


rsh52:
I think this is ready for review now, if I can get @ezraporter to review code (and the rest 😄 ) and @skadauke to review the [vignette](https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/dba13eddb5553df6918c23e40761d3a114e5f990/vignettes/articles/diving_deeper.Rmd#L176-L197) / [glossary](https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/dba13eddb5553df6918c23e40761d3a114e5f990/vignettes/glossary.Rmd#L228-L230) updates and the error message:

image

(note that the "Mixed Structure Instruments" hyperlink doesn't work yet until we've published this)

skadauke:
> I think this is ready for review now, if I can get @ezraporter to review code (and the rest 😄 ) and @skadauke to review the [vignette](https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/dba13eddb5553df6918c23e40761d3a114e5f990/vignettes/articles/diving_deeper.Rmd#L176-L197) / [glossary](https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/dba13eddb5553df6918c23e40761d3a114e5f990/vignettes/glossary.Rmd#L228-L230) updates and the error message:
> 
> image
> 
> (note that the "Mixed Structure Instruments" hyperlink doesn't work yet until we've published this)

I like the glossary, vignette, and error message updates! 👍

R/clean_redcap_long.R
```diff
@@ -337,3 +349,95 @@ distill_repeat_table_long <- function(form_name,
...
+#' \code{REDCapR::redcap_metadata_read()$data}
+#'
+#' @return
+#' Returns a \code{tibble} with list elements containing tidy dataframes. Users
+#' can access dataframes under the \code{redcap_data} column with reference to
+#' \code{form_name} and \code{structure} column details.
+#'
+#' @keywords internal
+
+convert_mixed_instrument <- function(db_data_long, db_metadata_long) {
```
ezraporter, line 372: We're losing data somewhere in here. When we convert the rows to have a value for `redcap_repeat_instrument` that was previously `NA` I think we're ending up throwing away the nonrepeating data from that row.

R/clean_redcap_long.R
```diff
@@ -22,19 +26,27 @@
...
 
   # Apply checkmate checks
   assert_data_frame(db_data_long)
   assert_data_frame(db_metadata_long)
 
   if (has_repeat_forms) {
-    check_repeat_and_nonrepeat(db_data_long)
+    if (enable_mixed_structure) {
+      db_data_long <- convert_mixed_instrument(db_data_long, db_metadata_long)
+      mixed_structure_forms <- get_mixed_structure_fields(db_data_long)
```
ezraporter, line 45: I'm a little confused about what calling this here after converting the data is doing. It looks like on the test data it's now flagging the nonrepeating fields?

ezraporter (changes requested):
This one is complicated and I'm having a little trouble wrapping my head around what's going on. I tried to add comments where I was getting confused.

It looks like we're losing data in the output though. The test redcap has 5 responses:
- Nonrepeat 1
- Nonrepeat 2
- Mixed Nonrepeat 1
- Mixed Repeat 1
- Mixed Repeat 2

Those first 2 should show up in `nonrepeat_form` but the `redcap_data` there has only "Nonrepeat 2":

```r
devtools::load_all()

data <- read_redcap(
  Sys.getenv("REDCAP_URI"),
  Sys.getenv("REDCAPTIDIER_MIXED_STRUCTURE_API"),
  enable_mixed_structure = TRUE
) |>
  extract_tibble("nonrepeat_form")

data$nonrepeat_1
#> [1] "Nonrepeat 2"
```
The data for `mixed_structure_form` looks okay.

R/clean_redcap_long.R
ezraporter: Another structure to consider that makes me wonder if our current approach actually generalizes:

|Event|Type|Nonrepeating Instruments|Repeating Instruments|
|-|-|-|-|
|Event 1 | Not Repeating | inst_1, inst_2, inst_3 | N/A|
|Event 2 | Repeat Instruments | inst_1 | inst_2, inst_3|

For the row in the data from `Event 1`, how would we be assigning `redcap_repeat_instrument`? In the test redcap it's easy because there's only data from one repeating instrument there but here we have data from 2 different repeating instruments.

R/read_redcap.R
```diff
@@ -75,7 +79,8 @@ read_redcap <- function(redcap_uri,
                         export_survey_fields = NULL,
                         export_data_access_groups = NULL,
                         suppress_redcapr_messages = TRUE,
-                        guess_max = .Machine$integer.max) {
+                        guess_max = .Machine$integer.max,
+                        enable_mixed_structure = FALSE) {
```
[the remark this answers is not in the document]
rsh52: Sure, I'm ambivalent so 👍

R/clean_redcap_long.R
```diff
@@ -55,7 +67,7 @@ clean_redcap_long <- function(db_data_long,
           linked_arms
         )
       ),
-      structure = "repeating"
+      structure = ifelse(has_mixed_structure_forms, "mixed", "repeating")
```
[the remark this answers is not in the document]
rsh52: Ah you're totally right, missed this in thinking for some reason that `clean_redcap_long` was being called iteratively in `read_redcap` but that's not right. Will fix.

R/clean_redcap_long.R
rsh52: I have to update this anyway per the atomic comment above, but on the line below it was needed to update `has_mixed_structure_forms`. I wanted to still handle the case where a user allows/enables mixed structure on a database that has repeating forms, but none that are actually mixed-structure. 

In this case `convert_mixed_instrument` won't have any effect, and `mixed_structure_forms()` will still return all `FALSE` values, leaving `has_mixed_structure_forms` `FALSE` down the line.

R/clean_redcap_long.R
rsh52: > We're losing data somewhere in here. When we convert the rows to have a value for redcap_repeat_instrument that was previously NA I think we're ending up throwing away the nonrepeating data from that row.

Ah, yes good catch. Will work on this one, seems less trivial than I was hoping. I'm not sure I follow the second comment but maybe fixing the first will address it.

ezraporter: Yeah I'd say address the first thing and then once that's working I can mess with the test redcap a bit to see if I can make the second thing I was saying more concrete

rsh52: First part addressed, let me know if this helps address the second.

R/read_redcap.R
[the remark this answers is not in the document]
rsh52: 🏳️ Tried [`options`](https://dgkf.github.io/options/) package which seemed ok but I wasn't a huge fan of some of the outputs. Couldn't find a lot online for how to get these set internal to the package in just base R or many examples in some other larger packages. Feel free to send documentation my way if you know of any good ones.

R/clean_redcap_long.R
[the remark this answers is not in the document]
rsh52: Should be resolved now.

R/read_redcap.R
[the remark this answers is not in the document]
ezraporter: I was thinking it would be as simple as changing the default for `enable_mixed_structure` to be `getOption("redcaptidier.allow.mixed.structure", FALSE)` and then just adding to the docs that you can set the option with `options(redcaptidier.allow.mixed.structure = TRUE)`. Maybe that's not kosher for some reason though.

`options` package looks interesting but maybe not necessary if we're not managing a whole bunch of options.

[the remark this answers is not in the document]
rsh52: Got it, overcomplicated it as usual. I haven't found any examples of this so unsure if kosher or not, but check out the most recent commit and see what you think. In practice it seems to be fine, I don't know if there's anything around this we need to do for testing or whatnot. Made a brief updated to the `?read_redcap()` params docs but let me know if you think it needs something more robust or a callout in the vignette.

R/clean_redcap_long.R
ezraporter: I made a new redcap that shows the issue. You have access to poke around but feel free to just use my token below. We'll delete this one later.

```r
devtools::load_all()

read_redcap(
  Sys.getenv("REDCAP_URI"),
  "0AE465FB49C411385DC8FE1C5A9638C0",
  allow_mixed_structure = TRUE
) |>
  extract_tibble("mixed_structure_form_2")
```

This one has 2 mixed structure instruments. The issue is that one row of the API export (the first row in this case) contains data from both mixed structure instruments. In `convert_mixed_instrument()` we assign a single instrument for each row of the export. It ends up just being the first one it sees. That results in the nonrepeating data from the second mixed instrument getting lost ("Mixed 2 Nonrepeat 1").

I haven't quite wrapped my head around this enough to suggest a fix but we're going to need to make things flexible enough that single rows of the export can get mapped to multiple repeating instruments.

rsh52: Ugh, alright I also need to wrap my head around this more. Thanks for the catch and explanation.

rsh52: Most recent commit: I _think_ this works (it seems to resolve your REDCap and maintain the old one, all tests and checks pass). 

The idea is that by moving the `convert_mixed_instrument()` function into the `distill_*()` function, we update `redcap_repeat_instrument` with each evaluation of the `map()` in `clean_redcap_long()`...

If I could post a GIF from Inception I would here. Before CRAN submission I almost want to ask the folks who submitted issues to test it.

ezraporter: Ah clever! That sounds like it makes sense but I'll take a closer look. Agree that once we get this merged we should ask the issue submitters to try out the dev version before submitting to CRAN

R/clean_redcap_long.R
```diff
@@ -86,6 +75,49 @@ clean_redcap_long <- function(db_data_long,
...
+    repeated_forms_tibble <- tibble(
+      redcap_form_name = repeated_forms,
+      redcap_data = map(
+        .data$redcap_form_name,
+        ~ distill_repeat_table_long(
+          .x,
+          db_data_long,
+          db_metadata_long,
+          linked_arms,
+          has_mixed_structure_forms = has_mixed_structure_forms,
```
ezraporter, line 110: I think this is fine but noting that one implication of this is that `convert_mixed_instrument()` gets run for every instrument even if it's actually repeating rather than mixed. It works because because `mixed_structure_ref` gets filtered to 0 rows for repeating instruments and the `for` loop in `convert_mixed_instrument()` doesn't run. We do still need to `filter()` `mixed_structure_ref` every time to find that out though which could be a performance hit. No need to optimize until it's a problem though

ezraporter (approved):
Okay I've 99% convinced myself this works and I actually think your solution is really good. I say we:
1. Run our timing benchmarks to make sure the latest change didn't mess things up. (Can we also add identifiers to the `microbenchmark_results.csv` so it's easy to line things up when we add new redcaps?)
2. Merge this and ask for the issue opener to test the dev version

R/clean_redcap_long.R
rsh52: Does it? The actual conversion is wrapped in a check on `has_mixed_structure_forms` so if `FALSE` (which it would be for the whole `map()` sequence) then `convert_mixed_instrument()` wouldn't get run unless I'm missing something.

https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/65d5baf88a5d916e22a385957b7dd5e304409607/R/clean_redcap_long.R#L315-L317

rsh52:
> Run our timing benchmarks to make sure the latest change didn't mess things up. (Can we also add identifiers to the microbenchmark_results.csv so it's easy to line things up when we add new redcaps?)

Updated run results and added column outputs for the database description and source (ouhsc / redcaptidier).

R/clean_redcap_long.R
ezraporter: Right but if you have mixed and repeating instruments then `has_mixed_structure_forms` is `TRUE` and it gets run for all the repeating instruments in addition to the mixed instruments.

ezraporter (approved):

Issue #178, opened 2024-03-08
[FEATURE] an option `raw_or_label = "haven_labelled"` for `read_redcap()`


larmarange:
# Feature Request Description

For some projects, our data workflow requires to import data from REDCap, make diverse transformations within R and then re-export data to several formats including Stata and SPSS using `haven` package.

For categorical variables, we need to keep the original codes used to collect data in REDCap while providing value labels. In `haven`, such data format is handled with the `haven_labelled` class, and several packages exist to deal with it, in particular the `labelled` package: https://larmarange.github.io/labelled/ (to be noted, this package is already suggested by REDCapTidieR).

For this purpose, it would be great to be able to import datasets from REDCap directly with the `haven_labelled` format, allowing to store both codes and value labels in the dataset.

# Proposed Solution

For that, it would be a great addition to have an option `raw_or_label = "haven_labelled"` in `read_redcap()` or alternatively a function `make_haven_labelled()` that could be applied on a super-tibble.

Technically:
1. data would have to be imported as `"raw"` with the current implementation of `read_redcap()`
2. the column `select_choices_or_calculations` in `redcap_metadata` will have to be parsed to identify and generate a list of value labels
3. then, such list could be easily applied to each tibble using `labelled::set_value_labels()` (through `.labels` argument)

In fact, the real piece of work is the point 2.

**Describe alternatives you've considered**

We are considering developing our own internal function. However, we have the feeling that such functionality could be useful for others and that it would be relevant to be integrated in an existing package.

## Additional Context

Please let us know if you think that such functionality would be relevant for `REDCapTidieR`. If yes, we would be happy to contribute and/or propose a PR.

### Checklist

- [x] The issue is atomic
- [x] The issue description is documented
- [x] The issue title describes the problem succinctly
- [x] Developers are assigned to the issue
- [x] Labels are assigned to the issue

larmarange:
Exploring the code of your package, it seems that most of the work is already done in the internal function `multi_choice_to_labels()`. It would therefore be releatively easy to have an option in this function to convert into an `haven_labelled` factor instead of a factor.

ezraporter:
Thanks for the feature request! This sounds generally useful so we plan on exploring this week and seeing what we can add. We'll be in touch soon!

larmarange:
Thank you.

larmarange:
I just tested. It worked perfectly. Thanks a lot.

PR #180, opened 2024-03-14
Add `haven` option to `raw_or_label`

ezraporter:
# Description
This PR modifies `read_redcap()` to let users specify `raw_or_label = "haven"` and have categorical fields converted to `haven_labelled` vectors instead of factors. I kept `labelled` in `Suggests` and added a check that the user has it installed if they specify `raw_or_label = "haven"`.

# Implementation considerations

`haven_labelled` vectors, unlike factors, preserve the underlying data values/types. The non-trivial part about this is that to apply the labels, the data types in the data must match the data types of the vector of values we read from the metadata. This is slightly tricky because there's no foolproof way to say "cast vector x to type of vector y" for arbitrary y. My approach was to basically do our best at casting using `readr`'s parsing functions and fall back to converting the underlying values to `chr` if all else fails.

See below for a concrete example if it's helpful.

# Proposed Changes 
- Add `"haven"` option to `read_redcap()` `raw_or_label` and update appropriate arg checks
- Refactor `multi_choice_to_labels()`
    - Now takes a `raw_or_label` argument from `read_redcap()`
    - Passes off logic for how categorical fields are handled to a `label_handler` function which will be one of `apply_labs_factor()` or `apply_labs_haven()`

# Worked Example

Suppose we have a redcap field coded like this:
|value|label|
|-|-|
|3|apple|
|5|orange|
|9|banana|

The `db_data` we get from redcap will have a field like this:

```r
 my_field
    <int>
        5
        3
        5
        9
```

where the `int` datatype was actually determined by `readr` since we don't allow users to pass data type specifications to `redcap_read_oneshot()`

The `db_metadata` will contain

```r
 field_name   select_choices_or_calculations         
 <chr>        <chr>                                  
 my_field     3, apple | 5, orange | 9, banana
```

where the values are stored as `chr` rather than `int`.

My implementation:
1. Checks the data type of `db_data$my_field` and finds it's `int`
2. Casts the labels read from `db_metadata`, `c(apple = "3", orange = "5", banana = "9")`, to `int` with `readr::parse_integer()`
3. Applies the labels with `labelled::set_value_labels()`

If step 2 had failed for some reason then it would cast `db_data$my_field` to `chr` and apply the labels to that.

### Issue Addressed
Related to #178

### PR Checklist

# Code Review

### Code Review Checklist


R/utils.R
```diff
@@ -485,15 +472,122 @@ multi_choice_to_labels <- function(db_data, db_metadata) {
...
+force_cast <- function(x, ptype) {
+  ptype <- vec_ptype(ptype)
+  if (is.logical(ptype)) {
+    out <- parse_logical(x)
+  } else if (is.integer(ptype)) {
+    out <- parse_integer(x)
+  } else if (is.numeric(ptype)) {
+    out <- parse_double(x)
+  } else if (is.Date(ptype)) {
+    out <- parse_date(x)
```
rsh52, line 578: I think we can address these Codecov warnings by adding the associated data types to tests in your new tests. I don't think it's useful to write a test for this function in particular since it's so basic, but you could do either.

https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/c58e78363fa246589a0bbfd1d05554c1164f9e01/tests/testthat/test-utils.R#L436

R/utils.R
```diff
@@ -485,15 +472,122 @@ multi_choice_to_labels <- function(db_data, db_metadata) {
...
   }
   db_data
 }
 
+#' @title
+#' Apply factor labels to a vector
+#'
+#' @param x a vector to label
+#' @param labels a named vector of labels in the format `c(value = label)`
+#' @param \dots unused
```
rsh52, line 491: Do we even need `...` as an option then? Is this just for future-proofing?

R/utils.R
```diff
@@ -485,15 +472,122 @@ multi_choice_to_labels <- function(db_data, db_metadata) {
...
+#' Assumes a check_installed() has been run for `labelled`. Since `haven` preserves the
+#' underlying data values we need to make sure the data type of the value options in the metadata matches
+#' the data type of the values in the actual data. This function accepts a prototype, usually a column
+#' from db_data, and uses `force_cast()` to do a best-effort casting of the value options in the metadata
+#' to the same data type as `ptype`. The fallback is to convert `x` and the value labels to character.
+#'
+#' @param x a vector to label
+#' @param labels a named vector of labels in the format `c(value = label)`
+#' @param ptype vector to serve as prototype for label values
+#' @param \dots unused
```
rsh52, line 516: Ditto above

R/utils.R
```diff
@@ -485,15 +472,122 @@ multi_choice_to_labels <- function(db_data, db_metadata) {
...
+  labels_cast <- withCallingHandlers(
+    {
+      force_cast(labels, ptype)
+    },
+    warning = function(w) {
+      cnd <<- w
+      cnd_muffle(w)
+    }
+  )
+  if (!is.null(attr(labels_cast, "problems"))) {
```
rsh52, line 537: Just a thought, should this be transitioned to a new `check_*()` function for organizational purposes?

rsh52 (approved):
Great work! Some minor comments, but I think the output looks like what's needed and there were some clever methods used here. 

I would next see if we can get the issuer to test this and get their feedback on if it meets expectations.

R/utils.R
ezraporter: In this case we do need it because this function gets called as `label_handler` which takes a `ptype` when `label_handler <- apply_labs_haven`. The `...` just allows that `ptype` to get passed through unused. I'll make the annotation clearer.

R/utils.R
ezraporter: In this case we actually don't need it since the parameters of `apply_labs_haven()` are a superset of `apply_labs_factor()`. I kept it around to mirror `apply_labs_factor` but we can remove

R/utils.R
ezraporter: I'm not opposed to this. My thought was that these checks are more logic that implements what `apply_labs_haven()` does rather than general data/arg checks like we have in the `check_*()` stuff. I worry that it will actually be less clear to separate this from the `withCallingHandlers()` above.

R/utils.R
ezraporter: Good call. Will do! Btw are those code coverage warnings getting added to the conversation new? I've never seen them before

rsh52: Unsure, I started noticing them in the past month or two. Relatively helpful for helping keep our % up

R/utils.R
rsh52: Fair enough, leave as is then

R/utils.R
ezraporter: Okay interesting finding (and I think why I skipped including those tests to start).

With `labelled` if you try to apply labels to vector that has no corresponding `val_labels<-` method it just returns the unlabelled vector:

```r
labels <- as.Date(c("The first day of 2023" = "2023-01-01"))

x <- labelled::set_value_labels(as.Date("2023-01-01"), .labels = labels)

x
#> [1] "2023-01-01:

class(x)
#> [1] "Date"
```

That means if someone has some crazy redcap where the value options are dates and they do `raw_or_label = "haven"`, their value labels won't get applied at all. BUT if they went ahead and implemented a `val_labels<-.Date` method then they would get applied. This is such a weird situation I don't think we should even care tbh. I added tests for `force_cast()` just to get rid of the coverage warnings.

rsh52: Huh, very interesting find. In that case yea, much better to just test `force_cast()`. If for _some_ reason someone encounters this then an issue can be opened. 

I figured we wouldn't test `force_cast()` since most of the options for it were already tested in the parent function but whatever appeases the CodeCov gods is fine.

larmarange: Value labels could be applied only to character or numeric vectors. It is not possible to add value labels to date vectors.

In fact, an error should have occured. I have open an issue on labelled website: https://github.com/larmarange/labelled/issues/156

PR #205, opened 2024-09-20
Missing Data Fix, Partial Keys Fix

rsh52:
# Description
This PR no longer implements `join_data_tibbles()`, that is being moved out into a new branch. Instead, this PR seeks to adress a few bugs, one minor (documented below), and one major where we are mistakenly reporting mixed data structure outputs.

I will be posting comments on the code itself below to explain some of the changes.

## The Missing Data Issue

Because we weren't making use of `redcap_event_instance` correctly, we were not accurately reporting all of the data associated with each REDCap event. See below from the Mixed Structure REDCap for reference:

image

And here is what currently comes out of the REDCapTidieR data tibbles from `main`:

<details>
  <summary><b>Current Output with Missing Data</b></summary>

```r
> sprtbl_mixed
# A REDCapTidieR Supertibble with 3 instruments
  redcap_form_name     redcap_form_label    redcap_data redcap_metadata   redcap_events structure data_rows data_cols
  <chr>                <chr>                <list>      <list>            <list>        <chr>         <int>     <int>
1 nonrepeat_form       Nonrepeat Form       <tibble>    <tibble [2 × 17]> <tibble>      nonrepea…         4         5
2 repeat_form          Repeat Form          <tibble>    <tibble [2 × 17]> <tibble>      mixed             2         5
3 mixed_structure_form Mixed Structure Form <tibble>    <tibble [2 × 17]> <tibble>      mixed             3         5
# ℹ 3 more variables: data_size <lbstr_by>, data_na_pct <formttbl>, form_complete_pct <formttbl>
> sprtbl_mixed$redcap_data
[[1]]
# A tibble: 4 × 5
  record_id redcap_event       redcap_event_instance nonrepeat_1 form_status_complete
      <dbl> <chr>                              <dbl> <chr>       <fct>               
1         1 non_repeating                         NA Nonrepeat 1 Incomplete          
2         1 repeating_separate                    NA Nonrepeat 2 Incomplete          
3         1 repeating_together                     1 A           Incomplete          
4         1 repeating_together                     2 B           Incomplete          

[[2]]
# A tibble: 2 × 5
  record_id redcap_event  redcap_form_instance repeat_1 form_status_complete
      <dbl> <chr>                        <dbl> <chr>    <fct>               
1         1 non_repeating                    1 Repeat 1 Incomplete          
2         1 non_repeating                    2 Repeat 2 Incomplete          

[[3]]
# A tibble: 3 × 5
[...]

PR #209, opened 2024-10-16
CRAN 1.2 Prep

rsh52:
# Description
This PR addresses notes and makes a few subtle changes to accommodate an oversight in repeat event definitions. Changes listed below:

- Several documentation links to external packages caused NOTE returns, these have been changes to backticks instead.
- One of the databases from OUHSC no longer works with the credential we have and is now skipped when running microbenchmark checks
- In `combine_checkboxes()` and `check_equal_col_summaries()` global variable NOTEs were returned for `.`, `.value`, and `.new_value`. The structural changes below have resolved this.
- The recent changes from #205 involving introducing `repeat_type` under the `repeat_events` col of the supertibble didn't account for longitudinal databases with no repeating architecture. The changes to `add_event_mapping` and `read_redcap()` should now appropriately accommodate longitudinal databases with and without repeating instances for forms and events.
- CRAN Comments prepped, NEWS prepped, [RHub run successfully](https://github.com/CHOP-CGTInformatics/REDCapTidieR/actions/runs/11371750216)
- Microbenchmark results updated

## Remaining TODOs and notes:

- Microbenchmark results seem to have gone up, not IMO alarmingly and IIRC this can at times just be the time of day or the fact a lot of our databases are getting larger. But worth noting.
- Some of the cli messages updated. I believe these are mostly related to message returns updating from REDCapR and how we mirror what it returns.
- Revdepcheck is deprecated, and we have the structure in place for using the new recheck workflow. However this is [currently failing](https://github.com/CHOP-CGTInformatics/REDCapTidieR/actions/runs/11371768665) because of `renv`.
   - I don't think this will be easily solvable since the steps used to build and install dependencies are under a [sub-workflow](https://github.com/r-devel/recheck/blob/v1/.github/workflows/recheck.yml) of `recheck`
   - IMO it might be time to bail on `renv`, in meeting with the ggplot2 devs it came up again that folks were surprised and frowned upon using renv in package development. Thoughts @ezraporter?
   - This is the last piece needed for the CRAN-Comments file
- In the future we may want to have some aspect of our pkgdown highlighting `combine_checkboxes()` but I don't think we need that now.

Thanks!


ezraporter:
> * IMO it might be time to bail on `renv`, in meeting with the ggplot2 devs it came up again that folks were surprised and frowned upon using renv in package development. Thoughts @ezraporter?

Okay I'm fine with this. We've clearly been outvoted and people are building tools that aren't compatible with `renv` in this area.

ezraporter (approved):

PR #225, opened 2025-07-04
add option in `combine_checkboxes()` to paste checkboxes when multiple selected

d-morrison:
# Description
Adds an option in `combine_checkboxes()` to concatenate checkbox values when multiple are selected.
Partially addresses #220; implements concatenation approach but not list-column approach.

**Screenshots**

Input:

```
> nonrepeat_data
# A tibble: 3 × 6
  study_id multi___1 multi___2 multi___3 single_checkbox___1 extra_data
     <dbl> <lgl>     <lgl>     <lgl>     <lgl>                    <dbl>
1        1 TRUE      FALSE     FALSE     TRUE                         1
2        2 TRUE      TRUE      FALSE     TRUE                         2
3        3 FALSE     FALSE     FALSE     FALSE                        3
```

Before:
```
# A tibble: 3 × 4
  study_id extra_data multi    single_checkbox
     <dbl>      <dbl> <fct>    <fct>          
1        1          1 Red      Green          
2        2          2 Multiple Green          
3        3          3 NA       NA      
```
After:
```
# A tibble: 3 × 4
  study_id extra_data multi       single_checkbox
     <dbl>      <dbl> <fct>       <fct>          
1        1          1 Red         Green          
2        2          2 Red, Yellow Green          
3        3          3 NA          NA        
```

### PR Checklist
Before submitting this PR, please check and verify below that the submission meets the below criteria:

- [x] New/revised functions have associated tests
- [x] New/revised functions that update downstream outputs have associated static testing files (`.RDS`) updated under `inst/testdata/create_test_data.R`
   - not sure if needed?
- [x] New/revised functions use appropriate naming conventions
- [x] New/revised functions don't repeat code
- [x] Code changes are less than **250** lines total
- [x] Issues linked to the PR using [GitHub's list of keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- [x] The appropriate reviewer is assigned to the PR
   - doesn't seem like I can do this?
- [x] The appropriate developers are assigned to the PR
   - doesn't seem like I can do this?
- [x] Pre-release package version incremented using `usethis::use_version()`

# Code Review
This section to be used by the reviewer and developers during Code Review after PR submission

### Code Review Checklist

- [ ] I checked that new files follow naming conventions and are in the right place
- [ ] I checked that documentation is complete, clear, and without typos
- [ ] I added/edited comments to explain "why" not "how"
- [ ] I checked that all new variable and function names follow naming conventions
[...]

PR #226, opened 2025-07-07
Combine Checkboxes Concatenation Refactor

rsh52:
This PR builds commits a slight refactor to #225 and moves the code into a branch that can access organization secrets so that Actions workflows pass.

I refactored the proposed changes to consolidate code together instead of in a larger if/else chunk. Other slight changes include:

- Package namespace documentation update (moved tidyr `unite` to NAMESPACE)
- Slight roxygen parameter documentation updates
- One update to a failing test not having access to the new `multi_value_sep` param

@d-morrison Apologies for moving this over to a new branch and PR, for security reasons we have some rules set up so that API tokens required for testing are left as organization secrets. Please review the most recent commit when you have a moment and let us know if you have any thoughts or concerns.


R/combine_checkboxes.R
```diff
@@ -272,19 +282,50 @@ replace_true <- function(col, col_name, metadata, raw_or_label) {
...
+              if (use_multi_value_label) multi_value_label else character(0)
+            ) %>%
+            union(
+              setdiff(
+                unique(.data[[.new_value]]),
+                values_fill
+              )
+            ) %>%
+            union(values_fill)
+        )
```
ezraporter, line 324: I would suggest returning a `character` vector rather than a `factor` if `multi_value_label` is `NULL`. In general `factor` levels should represent all possible data values, independent of what's actually in the data. That's clearly not feasible (or desirable) when every possible combination of values is an option so simply returning the `character` seems better and simplifies the code.

d-morrison: FWIW: for my use case, either format would be fine!

R/combine_checkboxes.R
```diff
@@ -272,19 +282,55 @@ replace_true <- function(col, col_name, metadata, raw_or_label) {
...
       )
     )
+
+  # If using multi_value_label, preserve factor levels
+  # Otherwise leave as character with concatenated values
+  if (use_multi_value_label) {
+    lvl_vec <- unique(c(
+      metadata[[raw_or_label]],
+      multi_value_label,
+      out[[.new_value]],
```
ezraporter, line 319: This _should be_ redundant because all values in the data need to come from one of the other options.

ezraporter (approved):
Awesome! Looks good to me!

R/combine_checkboxes.R
rsh52: You mean just `out[[.new_value]]`?

ezraporter: Yes!

PR #243, opened 2026-04-14
Repair malformed REDCap choice strings during label parsing

andrew-farrey:
# Description
Adds a conservative repair step before `parse_labels()` processes `select_choices_or_calculations` values from REDCap metadata.

The change is intended to improve resilience when metadata contain minor formatting defects that are still recoverable, such as repeated pipe separators or missing spacing after a separator before a numeric code. The repair step is intentionally narrow and preserves existing parser errors for strings that remain ambiguous or invalid after repair.

# Proposed Changes 

- Add a small preprocessing helper that repairs minor recoverable formatting defects in REDCap choice strings before parsing
- Update parse_labels() to use the repaired string while preserving its existing return contract for downstream helpers
- Add a regression test using placeholder values to confirm that repairable malformed strings parse successfully
- Preserve current error behavior for malformed strings that are not safely repairable
- Emit a warning when conservative formatting repair is applied before parsing

**Screenshots**
N/A

### Issue Addressed
`read_redcap()` would fail due to a missing space and an extra "|" delimiter in `select_choices_or_calculations` values that caused `parse_labels()` to fail. 

### PR Checklist

- [ ] New/revised functions that update downstream outputs have associated static testing files (`.RDS`) updated under `inst/testdata/create_test_data.R`
- [ ] Issues linked to the PR using [GitHub's list of keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)
- [ ] The appropriate reviewer is assigned to the PR
- [ ] The appropriate developers are assigned to the PR
- [ ] Pre-release package version incremented using `usethis::use_version()`

# Code Review

### Code Review Checklist

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214136483626380

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -201,6 +201,19 @@ parse_labels <- function(string, return_vector = FALSE, return_stripped_text_fla
     )
   }
 
+  repaired <- repair_redcap_choice_string(string)
+  string <- repaired$string
+
+  if (isTRUE(repaired$repaired)) {
+    cli_warn(
+      c(
+        "!" = "Applied conservative formatting repairs to a `select_choices_or_calculations` field before parsing.",
+        "i" = "This usually indicates malformed REDCap metadata such as repeated `|` separators or missing spacing after a separator."
+      ),
+      class = c("choice_string_repaired", "REDCapTidieR_cond")
+    )
+  }
+
   out <- string %>%
     strsplit("\\|") %>% # Split by "|"
     lapply(trimws) # Trim trailing and leading whitespace in list elements
@@ -272,6 +285,24 @@ parse_labels <- function(string, return_vector = FALSE, return_stripped_text_fla
   }
 
   out
+}
+
+repair_redcap_choice_string <- function(string) {
+  if (is.na(string) || !nzchar(trimws(string))) {
+    return(list(string = string, repaired = FALSE))
+  }
+
+  original <- string
+  repaired <- string
+
+  repaired <- gsub("\\|\\s*([0-9]+)\\s*,", "| \\1,", repaired, perl = TRUE)
+  repaired <- gsub("\\|\\s*\\|+", " | ", repaired, perl = TRUE)
+  repaired <- trimws(repaired)
+
+  list(
+    string = repaired,
+    repaired = !identical(original, repaired)
+  )
 }
 
 #' @title
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -121,6 +121,15 @@ test_that("parse_labels works", {
 
   invalid_string_1 <- "raw, label | that has | pipes but no other | commas"
   invalid_string_2 <- "raw, la|bel with pipe"
+  repairable_string_1 <- "803, Apple | 804, Banana | 805, Cherry |806, Dragon Fruit| | 807, Elderberry"
+  repairable_output_1 <- tibble::tribble(
+    ~raw, ~label,
+    "803", "Apple",
+    "804", "Banana",
+    "805", "Cherry",
+    "806", "Dragon Fruit",
+    "807", "Elderberry"
+  )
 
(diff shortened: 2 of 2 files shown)


andrew-farrey:
This PR was prompted by a real-world import failure in which `read_redcap()` encountered `select_choices_or_calculations` metadata containing small delimiter defects that were recoverable but not currently tolerated by `parse_labels()`.

ezraporter:
Thanks for the PR @andrew-farrey! Before moving forward can you clarify the intended scope of this fix.

The example string in your test case is malformed because of the empty entry between 806 and 807

```r
"803, Apple | 804, Banana | 805, Cherry |806, Dragon Fruit| | 807, Elderberry"
                                                          #^HERE
```

The missing whitespace around "806, Dragon Fruit" is handled already by existing code. You'll see for example that this parses as you expect:

```
"803, Apple | 804, Banana | 805, Cherry |806, Dragon Fruit| 807, Elderberry"
```

andrew-farrey:
Thanks, that’s helpful. You’re right that the missing whitespace is already handled. The actual failure case I was trying to address is the empty entry introduced by the extra separator. In projects where I've encountered this, a small number of orphan separators in select_choices_or_calculations cause parse_labels() to error, which in turn prevents read_redcap() from completing.

My goal here is only to recover from that narrow, mechanically repairable case, not to broaden parsing generally. I also meant “existing pre-parse cleaning step” more generally rather than a specific helper, so if there’s a better normalization point in or around parse_labels(), I’m happy to revise the PR.

ezraporter:
Got it! Can you provide a (possibly redacted) example of a `select_choices_or_calculations` that has caused this issue in real-world data?

My concern is that it's not possible to perfectly distinguish "orphan separators" from cases where the user has used the `|` or `,` character in a choice label. To my knowledge the former case isn't possible to create using the REDCap UI, while the latter case is possible and is the case that the error `parse_labels()` throws is designed for.

Given your description of this problem as co-occurring with missing whitespace I'm actually wondering if you're in the latter case and haven't realized it. But of course I don't know your data 😊

andrew-farrey:
Sure, the actual string that triggered the error came from a much longer list of toxicology panel answer choices. Here is a redacted example with the same structure as the metadata value that failed for me:

`203, Apricot | 204, Blueberry | 205, Clementine |206, Dalmatian| | 207, Elderflower`

And the corresponding segment from the real metadata had the same pattern:

`..., 805, [label A] |806, [label B]| | 807, [label C] | ...`

What I observed was:

- `|206,` by itself does not seem to be a problem (the missing whitespace)
- the failure appears to be triggered by the empty entry introduced by `| |`
- after collapsing that extra separator before parsing, the string becomes parseable and no error is thrown

The specific case I was trying to address is the accidental empty choice segment, not a general attempt to tolerate arbitrary `|` or `,` characters inside labels. 

I included handling for the missing-whitespace pattern because the current error message reports the original malformed string, so I did not realize that case was already effectively tolerated. I’m happy to revise the PR so it only handles that narrow pre-parse cleanup and leaves the existing whitespace behavior unchanged.

ezraporter:
The outstanding question is how you're proposing to differentiate empty entries from actually problematic inputs:

`203, Apricot | 204, Blueberry | 205, Citrus |ex. orange, clementine| | 207, Elderflower`

`203, \n | 204, ; | 205, | | 207, \t`

Our current heuristic is:

split on `|` -> trim whitespace -> check that every entry contains a comma separator.

Happy for you to propose an alternative heuristic in code by updating the PR but I'll probably continue discussing here whether it's something we want to accept.

If we decide there isn't an improved heuristic that we want to accept, we can also propose other workarounds that will get you unstuck. I acknowledge there likely isn't a perfect solution here because this issue stems from the imperfect way REDCap exposes value labels.

PR #244, opened 2026-05-22
Prep for 1.2.5

ezraporter:
This PR makes some minor updates to prepare for release of version 1.2.5:

- Formatted the repo with `air`, carrying over our `air` and `lintr` settings from the Trials Reporter
  - I removed `styler::style_pkg()` from `refresh.R` since `air` replaces it
  - I also reenabled `pipe_consistency_linter` with the setting `pipe = "auto"`. This checks that within each file there's only one type of pipe. We may want to think about converting to `|>` anyways at this point.
 - Resolved some warnings we were getting from `pkgdown::build_site()` about accessibility by adding alt text in a few spots
 - Updated `utility/test_creds.R` to not fail when some test REDCaps don't work
   - I noticed that all the `REDCapR` creds no longer work (403 error) but didn't investigate why

### PR Checklist

# Code Review

### Code Review Checklist


ezraporter:
Waiting on one final r-hub action to finish before sending the release to CRAN but we should be good to do that first thing Tuesday!

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+    "recommendations": [
```
rsh52, line 2: Should we be adding `.vscode/` to gitignore/Rbuildignore?

.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+    "[r]": {
+        "editor.formatOnSave": true,
```
rsh52, line 3: Same as above.

R/labelled.R
```diff
@@ -84,7 +84,8 @@ make_labelled <- function(supertbl, format_labels = NULL) {
   )
 
   ## Set some predefined labels for data fields that aren't in the metadata
-  data_labels <- c( # nolint: object_usage_linter
+  data_labels <- c(
+    # nolint: object_usage_linter
```
rsh52:
```suggestion
```

utility/refresh.R
```diff
@@ -15,6 +14,8 @@ devtools::check_man() #Should return NULL
 # Run as not CRAN to build full vignettes
 # devtools::check() sets NOT_CRAN = "true" by default but must set it manually
 # as needed for other devtools functions
+
+# NOTE: build_vignettes() was deprecated in devtools 2.5.0, build_article() recommended instead
 withr::with_envvar(
   new = c("NOT_CRAN" = "true"),
   devtools::build_vignettes()
```
rsh52: Per the note above, should we be using `build_article` here then?

tests/testthat/test-combine_checkboxes.R
```diff
@@ -1,39 +1,39 @@
...
-  "repeat___2", "checkbox", "1, A | 2, B | 3, C",
-  "repeat___3", "checkbox", "1, A | 2, B | 3, C"
+  ~"field_name" , ~"field_type" , ~"select_choices_or_calculations" ,
+  "study_id"    , "text"        , NA                                ,
+  "repeat___1"  , "checkbox"    , "1, A | 2, B | 3, C"              ,
+  "repeat___2"  , "checkbox"    , "1, A | 2, B | 3, C"              ,
+  "repeat___3"  , "checkbox"    , "1, A | 2, B | 3, C"
 )
 
 supertbl <- tibble::tribble(
```
rsh52, line 33: Can we add `tribble` and `first` (below) to the imports? I think maybe in the past we avoided this as it came up in the Check as an unused import, but I can't recall for certain.

rsh52 (changes requested):
Looking good! Thank you for working through it and applying the formatting. A few minor comments.

Re: The REDCapR APIs, this sounds like something we need to ask Will. I wouldn't hold up the submission for it, but we should ask him and cite the failures.

Re: The pipe, I think at this point that the base pipe is now old enough and popular enough that we can switch to it. I recall Stephan being open to that, or at least that it wasn't as much of a compatibility concern as it used to be. It will just lock us into a specific version of R for all users.

.vscode/extensions.json
ezraporter: Looks like it's already in the `.Rbuildignore`. `. gitignore` I could go either way. It's sort of personal settings that should be ignored but putting it in version control makes it more likely styling will get applied consistently.

rsh52: Ah, ok missed the build one. That's fine then, leave as is.

utility/refresh.R
ezraporter: The reason I didn't change it is because I'm not sure `build_article()` is a drop-in replacement but I think we should just bite the bullet.

tests/testthat/test-combine_checkboxes.R
ezraporter: Do you mean like in the `NAMESPACE`? The general convention I follow is to only include functions you need for package code (`R/`) but I know longer remember where that comes from. It sort of complies with the advice here but we don't follow it in general: https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-imports

ezraporter:
> Re: The REDCapR APIs, this sounds like something we need to ask Will. I wouldn't hold up the submission for it, but we should ask him and cite the failures.

Looks like he actually updated the source file with new credentials: https://github.com/OuhscBbmc/REDCapR/blob/5f40ab5fafc128ae4525a8cad4bf1fb938e449bf/inst/misc/example.credentials

Seeing if those work and if not I'll ping him

> Re: The pipe, I think at this point that the base pipe is now old enough and popular enough that we can switch to it. I recall Stephan being open to that, or at least that it wasn't as much of a compatibility concern as it used to be. It will just lock us into a specific version of R for all users.

Added an issue! #245

ezraporter:
> Looks like he actually updated the source file with new credentials: https://github.com/OuhscBbmc/REDCapR/blob/5f40ab5fafc128ae4525a8cad4bf1fb938e449bf/inst/misc/example.credentials

Worked! Updated in the last commit. I think we're all good so re-requested review

rsh52 (approved):
Great thank you! Looks like on a quick inspection between current and past microbenchmark times that they all went slightly up.. But could be any number of reasons including REDCap changes or time of day these were run.

PR #246, opened 2026-06-08
Determine instrument structure from project metadata

### LINKED ISSUE #234: [BUG] `read_redcap()` treats repeating instruments as nonrepeating if no data is present
# Expected Behavior
`read_redcap()` should correctly return a repeating instrument with zero rows when no data exists

# Current Behavior
`read_redcap()` returns with a non-repeating structure and a single row of `NA`s for all records

## How to Reproduce the Bug:
```r
db <- read_redcap(Sys.getenv("REDCAP_URI"), Sys.getenv("REDCAPTIDIER_NO_DATA_API"))

db

# Second instrument is repeating but gets structure of non-repeating 

#># A REDCapTidieR Supertibble with 2 instruments
#>  redcap_form_name redcap_form_label redcap_data redcap_metadata   structure data_rows data_cols data_size
#>  <chr>            <chr>             <list>      <list>            <chr>         <int>     <int> <lbstr_b>
#>1 nonrepeating     nonrepeating      <tibble>    <tibble [2 × 17]> nonrepea…         1         3 1.79 kB  
#>2 repeating        repeating         <tibble>    <tibble [2 × 17]> nonrepea…         1         3 1.74 kB

# Data includes blank response for record 1 that was never recorded
db |>
  extract_tibble("repeating")

#># A tibble: 1 × 3
#>  record_id repeating form_status_complete
#>      <dbl> <lgl>     <fct>               
#>1         1 NA        NA  
```

## Additional info

This bug was discovered while investigating #232. The fix isn't straightforward because we currently identify repeated forms based on their presence in `db_data$redcap_repeat_instrument`. Forms with no data will not appear so we'd need a different method of identifying repeating forms.

https://github.com/CHOP-CGTInformatics/REDCapTidieR/blob/b1189f2018996164c5d754fe62396aa41c5c0d58/R/clean_redcap.R#L35-L38

### Checklist

- [x] The issue is atomic
- [ ] Developers are assigned to the issue

ezraporter:
# Description
This PR updates `read_redcap()` to determine the structure of instruments using `REDCapR`'s new [`redcap_instrument_repeating()`](https://ouhscbbmc.github.io/REDCapR/reference/redcap_instrument_repeating.html). This is a fairly significant change since previously `REDCapTidieR` inferred structure from the data itself, causing the reported structure to change depending on whether data is completed. Since `redcap_instrument_repeating()` is not yet on CRAN we'll have to wait before doing a CRAN release but users should still be able to install the dev version from GitHub if we want to merge before a new `REDCapR` release.

# Proposed Changes 
- Updated core login in `read_redcap()` to create a `form_structure` `tibble` that gets passed down to `clean_redcap_*()` to inform which forms are nonrepeating/repeating/mixed
- Added new helper functions, `structure_from_repeating()`, `structure_from_events_and_repeating()`, and `add_form_event_structure()` to control determining structure to the form-event level and subsequently the form level.
- Updated `link_arms()` to join in the form-event repeating structure. That also let's us simplify `add_event_mapping()` since all the data we need is in `linked_arms` now.
- Updated `clean_redcap_*()` to accept the new `form_structure()` and replace instances where repeating/nonrepeating structure were inferred from the data
- Added additional test cases to `test-read_redcap` to ensure that the right structure is produced with and without complete data.
  - You'll have to add a new `REDCAPTIDIER_NO_DATA_API` key to run tests locally

### Issue Addressed
closes #234 

### PR Checklist

- [ ] Code changes are less than **250** lines total

# Code Review

### Code Review Checklist


R/read_redcap.R
```diff
@@ -672,3 +644,167 @@ update_dag_cols <- function(data, dag_data, raw_or_label) {
...
+  try_redcapr(
+    {
+      redcap_event_instruments(
+        redcap_uri = redcap_uri,
+        token = token,
+        arms = NULL, # get all arms
+        verbose = !suppress_redcapr_messages
+      )
+    },
+    call = if (is.null(call)) caller_env() else call
```
rsh52, line 716: Why is it necessary for this one to be different versus the `call` in `pull_instrument_repeating` above?

R/read_redcap.R
```diff
@@ -672,3 +644,167 @@ update_dag_cols <- function(data, dag_data, raw_or_label) {
...
+  db_instrument_repeating,
+  db_event_instruments
+) {
+  new_form_structure <- db_event_instruments %>%
+    add_form_event_structure(db_instrument_repeating) %>%
+    summarise(new_structure = structure_to_form_level(.data$repeat_structure), .by = "form")
+
+  form_structure %>%
+    left_join(new_form_structure, by = c("redcap_form_name" = "form")) %>%
+    mutate(structure = coalesce(.data$new_structure, .data$structure), .keep = "unused")
```
rsh52: Is `coalesce` needed here or can you just do `structure = .data$new_structure, .keep = "unused"`? 

`keep = "unused"` is neat, I've only ever used "none" and forgot there were other options.

R/read_redcap.R
```diff
@@ -672,3 +644,167 @@ update_dag_cols <- function(data, dag_data, raw_or_label) {
...
+#' @keywords internal
+add_form_event_structure <- function(data, db_instrument_repeating) {
+  db_instrument_repeating <- db_instrument_repeating %>%
+    mutate(repeat_structure = if_else(is.na(.data$form), "repeat_together", "repeat_separate"))
+
+  repeating_events <- db_instrument_repeating %>%
+    filter(is.na(.data$form)) %>%
+    select(
+      "unique_event_name",
+      custom_form_label_event = "custom_form_label",
```
rsh52, line 769: How come we need `custom_form_label_event` separate from `custom_form_label_form` in the next `select()` when they're both just getting `coalesce`d back under `custom_form_label` later anyway? 

I'm probably missing something, but if we don't need the rename/coalesce then we could just add `custom_form_label` into the `by` of the second `left_join`.

R/utils.R
```diff
@@ -119,10 +119,12 @@ create_repeat_instance_vars <- function(db_data) {
 #' @param token The REDCap API token
 #' @param suppress_redcapr_messages A logical to control whether to suppress messages
 #' from REDCapR API calls. Default `TRUE`.
+#' @param db_event_instruments result of [REDCapR::redcap_event_instruments()], retrieved from REDCap
```
rsh52: Maybe we just need to add that it will be pulled in the event its not supplied already upstream.

R/clean_redcap.R
```diff
@@ -18,24 +20,18 @@
 #'
 #' @keywords internal
 
-clean_redcap <- function(db_data, db_metadata) {
+clean_redcap <- function(db_data, db_metadata, form_structure) {
```
rsh52, line 23: `form_structure` having the side effect of simplifying these functions is nice 👍

rsh52 (changes requested):
Excellent work! I think this is a great up-front re-structure and a much more valid way of determining the form structure of the REDCap EDCs. A couple of small clarifications/requests but nothing major.

I'm also getting an error when running check/test locally for one snapshot test and wondering if you got it too:

> Failure ([test-write.R:371:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): Combining skimr, labelled, and xlsx returns expected snapshot

For a future release, I think it would be good to lump this large change in with a pipe update for CRAN whenever that comes about.

R/read_redcap.R
ezraporter: Only because this function gets called within `link_arms`. We want `call` to trace back to `read_redcap()` so if you're calling it directly in `read_redcap()` you can use the default. On second thought I'll add this to `pull_instrument_repeating()` so they're consistent.

rsh52: Ok cool, thank you for explaining!

R/read_redcap.R
ezraporter: Just doing this so `custom_form_label` isn't duplicated. There may be a better way? I don't think adding to the `by` in the second join will work. After the first join `custom_form_label` is `NA` for all the relevant rows in the next join so adding it would cause them not to join.

R/read_redcap.R
ezraporter: The `coalesce()` is handling the annoying but possible-in-REDCap edge case where a form isn't linked to any events. I'm actually not sure this works quite right because in that case the "nonrepeating" would get carried through which isn't quite right. Let me think about this one a bit more.

R/read_redcap.R
rsh52: Ah ok, since there weren't custom labels in the sample database I suppose this wasn't clear to me as everything was `NA`. Feel free to leave as is, it's not harming anything but it wasn't apparent why it was necessary.

R/read_redcap.R
rsh52: Hm yea, I'm wondering if there's a developer-followable middleground of saying "undetermined" or just not assigning a value at start, but have explicit logic that turns to "nonrepeating" instead of having it be the default assumption?

ezraporter: Implemented in the last commit. Based on our discussion the behavior is: warn if any form-events are marked repeating but not listed in the form-event mapping and report then as nonrepeating consistent with the Repeating instruments and events UI

ezraporter:
Re-requesting review since I think I've covered everything. I actually found a larger bug in how I was handling longitudinal nonrepeating redcaps and added additional `read_redcap()` tests that would have caught that.

> I'm also getting an error when running check/test locally for one snapshot test and wondering if you got it too:
> 
> > Failure (test-write.R:371:3): Combining skimr, labelled, and xlsx returns expected snapshot

I'm not seeing this one but I wonder if it's related to package versions. I usually update all my packages to the latest versions with `devtools::install_deps()` when jumping back to REDCapTidieR.

R/checks.R
```diff
@@ -950,3 +950,33 @@ check_equal_col_summaries <- function(data, col1, col2, call = caller_env()) {
...
+    pull("form") %>%
+    unique()
+
+  if (length(unmatched_forms) == 0) {
+    return()
+  }
+
+  msg <- c(
+    `!` = "The {.code {unmatched_forms}} form{?s} ha{?s/ve} repeating event information but {?is/are} not linked to any events. {.code structure} will be {.code nonrepeating}.", # nolint: line_length_linter
+    i = "This can occur when a previously previously repeating form was unlinked from all event."
```
rsh52: "This can occur when a previously ~previously~ repeating form was unlinked from all event**s**." 

Drop the extra "previously", pluralize "event".

Unsure why but suggesting a code change isn't working for me.

rsh52 (approved):
One small change to the message below. Otherwise think this is looking good! Nice catch and thanks for adding the additional tests.

The failing test worked for me this time, unsure what changed tbh!

R/checks.R
ezraporter: Ah thanks for catching!

